PR #3, opened 2015-02-15
First milestone, working with Travis.

abelsiqueira:


dpo:
I've incorporated your changes related to the Julia interface. The rest is tricky because there are many commits, some are duplicated, some undo things that others did, and many modify multiple files. Would you mind opening PRs to submit the following separately:

- the medium interface
- the raw interface
- anything related to Travis
- any other updates you'd like to merge.

I could simply copy the files, but I'd rather they retain your name on them. For the raw interface, could you put your Python script in some `aux` or `util` folder?

For now, the finalizer is still called `cutest_finalize` (not `terminate`) and it's still in `CUTEst.jl` (not in `julia_interface.jl`). The reason is that it seems the constructor and destructor should stay together, and Julia associates finalizers with `finalizer()`. However, in the raw interface, we should clearly use the name `terminate`.

I've also decided to not change all `int32`s to `Cint`s and all `float64`s to `Cdouble`s. You're right that Julia considers them the same, but just for clarity, I'd rather only use `Cint` and `Cdouble` to indicate explicitly that we're doing something that has to do with calling a C function.

Finally, I made a small change to `objgrad()`, where the objective value wasn't returned as a scalar.

Many thanks again for all your work!

abelsiqueira:
Hey, just passing to say that I read, and will try to work on that later this week.

abelsiqueira:
My class was canceled, so I taking a quick look around.
Your commit e9d9edb is not doing what was intended. Running CSETUP twice "does not work". See Issue #4.

I'll start cherry-picking the commits for the different pull-requests, but I'll use 7beb087 as base.

dpo:
I know. I just haven't addressed that problem yet. I was hoping to fix CUTEst to keep this history clean.

abelsiqueira:
I previously mentioned that I was gonna use 7beb087 as base to the new pull request, but working on the code, I noticed that e9d9edb can be used as a substitute for `initialized` on the Model. So I rebase what I was doing, and made the first pull request.
Also, I'm closing this.

PR #5, opened 2015-02-27
Create Travis and changes the minimum in code to make it work

abelsiqueira:
Creates `.travis.yml` and a file to install `CUTEst` on Travis. Change very little in the code to make Travis work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: julia
+os:
+  - linux
+  - osx
+julia:
+  - release
+  - nightly
+before_install:
+  - ./before_install.sh
+script:
+  - source $TRAVIS_BUILD_DIR/src_vars
+  - julia -e 'Pkg.clone(pwd()); Pkg.build("CUTEst"); Pkg.test("CUTEst", coverage=true)'
+after_success:
+  - julia -e 'cd(Pkg.dir("CUTEst")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
+matrix:
+  allow_failures:
+    - os: osx
--- a/before_install.sh
+++ b/before_install.sh
@@ -21,24 +21,18 @@ if [ "$TRAVIS_OS_NAME" == "linux" ]; then
   ln -s $(which gfortran-$gccver) $brewdir/bin/gfortran
 
   PATH=$brewdir/bin/:$PATH
-  LD_LIBRARY_PATH=$brewdir/lib/:$LD_LIBRARY_PATH
+  sudo ln -s /usr/lib/gcc/x86_64-linux-gnu/4.6/libgfortran.so /usr/local/lib
 fi
 
 brew tap dpo/cutest
 brew install cutest --HEAD
 brew install mastsif --HEAD
 echo "export PATH=$PATH" > $TRAVIS_BUILD_DIR/src_vars
-echo "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH" >> $TRAVIS_BUILD_DIR/src_vars
 for f in archdefs mastsif cutest sifdecode
 do
   cat $(brew --prefix $f)/$f.bashrc >> $TRAVIS_BUILD_DIR/src_vars
 done
 
-#Workaround for LD_LIBRARY_PATH
-if [ "$TRAVIS_OS_NAME" == "linux" ]; then
-  sudo ln -s $(brew --prefix cutest)/lib/libcutest.so /usr/local/lib
-  sudo ln -s /usr/lib/gcc/x86_64-linux-gnu/4.6/libgfortran.so /usr/local/lib
-fi
 
 julia -e 'Pkg.clone("https://github.com/optimizers/NLP.jl.git")'
 
--- a/before_install.sh
+++ b/before_install.sh
@@ -4,22 +4,40 @@ set -e
 
 if [ "$TRAVIS_OS_NAME" == "linux" ]; then
   sudo apt-get update -qq -y
-  sudo apt-get install -y gfortran
-  git clone https://github.com/Homebrew/linuxbrew.git $TRAVIS_BUILD_DIR/linuxbrew
-  PATH=$PATH:$TRAVIS_BUILD_DIR/linuxbrew/bin/
-  brew tap dpo/cutest
-  brew install cutest --HEAD
-  brew install mastsif --HEAD
-  echo "export PATH=$PATH" > $TRAVIS_BUILD_DIR/src_vars
-  for f in archdefs mastsif cutest sifdecode
-  do
-    cat $(brew --prefix $f)/$f.bashrc >> $TRAVIS_BUILD_DIR/src_vars
-  done
-  sudo ln -s $(brew --prefix cutest)/lib/libcutest.so /usr/local/lib/
-  sudo ln -s /usr/lib/gcc/x86_64-linux-gnu/4.6/libgfortran.so /usr/local/lib/
-else
-  # Implement before_install for OSX here
-  echo "No support for OSX" yet
+
+  #install LinuxBrew
+  sudo apt-get install build-essential curl git m4 ruby texinfo libbz2-dev libcurl4-openssl-dev libexpat-dev libncurses-dev zlib1g-dev
+  echo -ne '\n' | ruby -e "$(wget -O- https://raw.github.com/Homebrew/linuxbrew/go/install)"
+  brewdir=$HOME/.linuxbrew
+
+  #Symlink GCC to avoid installing Homebrew GCC
+  gccver=$(gcc -dumpversion | cut -d. -f1,2)
+  ln -s $(which gcc) $brewdir/bin/gcc-$gccver
+  ln -s $(which g++) $brewdir/bin/g++-$(g++ -dumpversion |cut -d. -f1,2)
+
+  #Install GFORTRAN and add to visible location
+  sudo apt-get -y install gfortran-$gccver
+  ln -s $(which gfortran-$gccver) $brewdir/bin/gfortran-$gccver
+  ln -s $(which gfortran-$gccver) $brewdir/bin/gfortran
+
+  PATH=$brewdir/bin/:$PATH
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: julia
+os:
+  - linux
+  - osx
```
dpo: I just requested multi-os builds. It might be available tomorrow.

before_install.sh
```diff
@@ -0,0 +1,27 @@
...
+  git clone https://github.com/Homebrew/linuxbrew.git $TRAVIS_BUILD_DIR/linuxbrew
+  PATH=$PATH:$TRAVIS_BUILD_DIR/linuxbrew/bin/
+  brew tap dpo/cutest
+  brew install cutest --HEAD
+  brew install mastsif --HEAD
+  echo "export PATH=$PATH" > $TRAVIS_BUILD_DIR/src_vars
+  for f in archdefs mastsif cutest sifdecode
+  do
+    cat $(brew --prefix $f)/$f.bashrc >> $TRAVIS_BUILD_DIR/src_vars
+  done
```
dpo: Lines 10-17 should apply to both Linux and OSX. On Linux you also need to add `$TRAVIS_BUILD_DIR/linuxbrew/lib` to the front of `LD_LIBRARY_PATH`.

before_install.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -e
+
+if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+  sudo apt-get update -qq -y
+  sudo apt-get install -y gfortran
+  git clone https://github.com/Homebrew/linuxbrew.git $TRAVIS_BUILD_DIR/linuxbrew
+  PATH=$PATH:$TRAVIS_BUILD_DIR/linuxbrew/bin/
```
dpo: Here `linuxbrew/bin` should be in the front.

before_install.sh
```diff
@@ -0,0 +1,27 @@
...
+  brew tap dpo/cutest
+  brew install cutest --HEAD
+  brew install mastsif --HEAD
+  echo "export PATH=$PATH" > $TRAVIS_BUILD_DIR/src_vars
+  for f in archdefs mastsif cutest sifdecode
+  do
+    cat $(brew --prefix $f)/$f.bashrc >> $TRAVIS_BUILD_DIR/src_vars
+  done
+  sudo ln -s $(brew --prefix cutest)/lib/libcutest.so /usr/local/lib/
```
dpo: Have a look at https://github.com/dpo/homebrew-cutest/blob/master/setup_travis_linux.sh to avoid building `gcc` on Linux.

before_install.sh
abelsiqueira: I actually wanted to do this and remove line 18. But then it can't find `-lcutest`.

https://travis-ci.org/abelsiqueira/CUTEst.jl/jobs/52410166#L222

Did I miss something?

dpo: Did you update `LD_LIBRARY_PATH` in that run? I don't see it. Also you might want to `brew test -v sifdecoder` and `brew test -v cutest` for good measure. Remember that they build `HEAD`, not a release.

src/CUTEst.jl
```diff
@@ -53,7 +53,7 @@ function sifdecoder(name :: ASCIIString)
   libname = "lib$pname";
   run(`sifdecoder $name`);
   run(`gfortran -c -fPIC ELFUN.f EXTER.f GROUP.f RANGE.f`);
-  run(`gfortran $sh_flags -o $libname.$soname -lcutest ELFUN.o EXTER.o GROUP.o RANGE.o`);
+  run(`ld $sh_flags -o $libname.$soname ELFUN.o EXTER.o GROUP.o RANGE.o -lcutest -lgfortran`);
```
dpo: Why is this necessary?

before_install.sh
abelsiqueira: This is the change: b5e14e5

src/CUTEst.jl
abelsiqueira: I have this error:
````
ERROR: ccall: could not find function fortran_open_ in library libBYRDSPHR
````

before_install.sh
dpo: Ah I think on TravisCI, environment variables that are set in a script disappear when the script exits. You have to set them in `.travis.yml`.

abelsiqueira: That is true, but that's why I send them to `src_vars` and in the next step I source `src_vars`.
Here is a debugging showing that this works:
https://travis-ci.org/abelsiqueira/CUTEst.jl/jobs/52455490#L213

abelsiqueira:
I updated the `before_install.sh` to use your changes.
The cutest is still missing if I don't `ln` to `/usr/local/lib` or similar. So that part remains.

before_install.sh
dpo: You're right. It seems that it's Travis that messes up.

src/CUTEst.jl
dpo: And if you move `-lcutest` to the end, it fails as well?

On OSX it should be `libtool`, but please try `gfortran` first.

abelsiqueira: Still fails:

dpo: The OSX builds succeeded.

dpo:
[Here](https://travis-ci.org/optimizers/CUTEst.jl/jobs/52505181#L524) I tried to add `~/.linuxbrew/lib` to the library dir directly in Julia, but that didn't work either. Something is just wrong with their Linux setup.

abelsiqueira:
I'm still trying to make use `gfortran` with linux, without success.
In [this](https://travis-ci.org/abelsiqueira/CUTEst.jl/builds/52500006) build, I made osx and linux work, using `gfortran` for osx as usual, and `ld` for linux. It's ugly, but it works.
I just saw you issue. Let's hope someone can find a better solution.

src/CUTEst.jl
dpo: You're not having this problem locally, are you?

abelsiqueira: In my arch linux no, but I tested with a Ubuntu, and the problem exists. Note that I'm using `libcutest.so` in `/usr/local/lib`.
In Ubuntu, when I link with gfortran, and then use `ldd`, I get "statically linked". When using `ld` I receive all the expected information.

dpo:
Ok so [this](https://travis-ci.org/optimizers/CUTEst.jl/builds/52548038) works. Here's what I changed:

* [Give the location of the library](https://github.com/optimizers/CUTEst.jl/commit/f9179e7eebacb8976bb302b045a680c5456a37e4)
* [Use `ld` on Linux](https://github.com/optimizers/CUTEst.jl/commit/16f5fdb91cd19e3c6099c9a8b1f1e513bf9903bf)

If you installed CUTEst with Homebrew/Linuxbrew, you'll have to
````
brew uninstall cutest
brew update
brew install cutest
````
because I changed the value of the `CUTEST` env var that the Julia code now uses. That was necessary on OSX.

dpo:
Does the resolution seem reasonable to you?

My suggestion is that we try to get (nearly) 100% coverage with the Julia interface before merging the medium and expert interfaces.

Thanks, as always!

abelsiqueira:
> Does the resolution seem reasonable to you?
>   - Give the location of the library
Very reasonable.
>   - Use ld on Linux
This is what I did, so yes.

I had already done the change on the linker (on another branch). Now I cherry-picked your lib location and updated travis. I will now push these changes to this pull request.

About the (nearly 100%) coverage on Julia, it seems reasonable. I almost had it already, but it was hidden under all the very low coverage of the other packages.

I also wanted to update the test to verify if the values obtained were correct. I defined manually a `hs32.jl` and tested after each cutest function call if the values were close enough. That's how I verified that calling csetup twice was not working.
Note that, since we can't call csetup twice, we can't have 100% coverage until we can open both constrained and unconstrained problems in the same test.

abelsiqueira:
I'm just passing to say that I've been working on the coverage and testing on the julia interface, and it is now at 96.49% coverage for `julia_interface.jl`.

I'm working on a different branch for that, but I didn't create a pull request yet because it if you accept both, it looks like they won't be fast-forwarded (as far as I remember).

The commits can be seen at the [network](https://github.com/optimizers/CUTEst.jl/network).

abelsiqueira:
I saw that you accepted the commits about `hess` and `hess_coord`, but not this pull request.
Is something missing here? Do you intend to fix that cutest issue before accepting this?

dpo:
It's just that I've been working with https://github.com/optimizers/CUTEst.jl/commit/ace07054a6b2f73d387fc020bc208630281fab77 to debug the Travis issue we were having last week. Also my repo can't have OSX enabled right now because Travis is currently running at full capacity.

abelsiqueira:
You mean `-lcutest` missing, or the linker?

About the cherry-picking, I would prefer merging branches into master, rather than pushing directly.
It is, of course, your repo, but I'm suggesting (if you still intend to give me writing rights) that we don't push to `master` directly, only create pull requests, and the other person accepts it.
This would allow clean merging, and if we disagree on something, that won't be on `master` before discussing (like my first pull request, or the first version of `bug/hess-symm`).

dpo:
Oh I totally agree. Locally, I have a `develop` branch. I should push it. I've been cherry-picking to keep changes as clean as possible. Several things were combined together in your branches and PRs. That's why I cherry-picked. I was careful to leave your name on your changes. I agree with you that we should keep master pristine. I've been using the `git-flow` model.

I think the Travis issues are resolved now. Are there any major differences between your scripts and mine?

abelsiqueira:
`git-flow` seems like a great model, but I'm suggesting to never pushing directly to `develop` (that wasn't clear for me).
So
  - `develop` only walks when accepting pull requests
  - `master` only walks when we decide to create a new release

I don't think there are many differences.

dpo:
Sounds good. I'll push my current `develop` and we'll work with PRs from here on.

dpo:
If you don't mind, I'll close this and create a PR from the `feature/travis` branch because the Travis folks can't enable OSX builds for this repo at this time.

abelsiqueira:
Ok

PR #11, opened 2015-03-05
Improve testing and coverage

abelsiqueira:
Test julia interface functions against expected values.
Change layout of testing, so that the next tests will be easier to make.


dpo:
Looks great, thanks!

PR #12, opened 2015-03-05
Create unconstrained testing against HS4

abelsiqueira:
HS4 is unconstrained, and the test code is the same.
Note that, currently, we can't call these two tests in the same julia instance, so the `Pkg.test` only runs with HS32, and then we create another julia instance to run `runtests_unc.jl`. This means no increase in coverage.

### DIFF
--- a/test/hs4.jl
+++ b/test/hs4.jl
@@ -1,0 +1,7 @@
+# HS4 definitions for testing purposes
+nvar = 2
+ncon = 0
+f(x) = (x[1]+1)^3/3 + x[2]
+g(x) = [(x[1]+1)^2; 1]
+H(x) = [2*(x[1]+1) 0; 0 0]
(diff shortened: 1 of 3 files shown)


test/hs4.jl
```diff
@@ -0,0 +1,7 @@
+# HS4 definitions for testing purposes
+nvar = 2
+ncon = 0
+f(x) = (x[1]+1)^3/3 + x[2]
+g(x) = [(x[1]+1)^2; 1]
+H(x) = [2*(x[1]+1) 0; 0 0]
+x0 = [1.125; 0.125]
```
dpo: Why is it necessary to also define `x0` here? Shouldn't we always use `x0 = nlp.meta.x0`?

abelsiqueira: True, my mistake. Will fix this and open another PR.

PR #17, opened 2015-03-13
Core interface

abelsiqueira:
This bundle contains
- The script to generate, and the generated, core interface
- Tests for a large part of the core interface
- Change in `CUTEst.jl` to use the core interface instead of the `ccall`.

Some constrained functions were not tested, specially the finite element related.
Most unconstrained functions were not tested, as they were not going to improve coverage.


dpo:
Thanks a lot for all this! Sorry I was away for a while.

dpo:
Something might have changed in Julia nightly:
https://travis-ci.org/optimizers/CUTEst.jl/jobs/56270657#L552

PR #18, opened 2015-03-30
Fix for DL_LOAD_PATH

abelsiqueira:
DL_LOAD_PATH has been moved from Base to Base.Libdl.
Compat provides de backwards compatibility


dpo:
Thanks!

PR #19, opened 2015-04-22
Add sudo: require, as instructed on travis

abelsiqueira:
[Why `sudo: require`?](http://blog.travis-ci.com/2015-03-31-docker-default-on-the-way/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ julia:
   - release
   - nightly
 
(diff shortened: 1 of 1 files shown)


dpo:
Is it `require` or `required`?

abelsiqueira:
Apparently it should be `required`, but this is working. Nevertheless, I'll fix this.

PR #21, opened 2015-05-08
Fix type warnings and & in objcons ccall

abelsiqueira:
Remove some warnings, and fix a `ccall`.

### DIFF
--- a/src/julia_interface.jl
+++ b/src/julia_interface.jl
@@ -9,7 +9,7 @@ function objcons(nlp :: CUTEstModel, x :: Array{Float64,1})
   if ncon > 0
     @eval ccall((:cutest_cfn_, $(nlp.libname)), Void,
                 (Ptr{Int32}, Ptr{Int32}, Ptr{Int32}, Ptr{Float64}, Ptr{Float64}, Ptr{Float64}),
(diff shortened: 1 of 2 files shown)


src/julia_interface.jl
```diff
@@ -9,7 +9,7 @@ function objcons(nlp :: CUTEstModel, x :: Array{Float64,1})
   if ncon > 0
     @eval ccall((:cutest_cfn_, $(nlp.libname)), Void,
                 (Ptr{Int32}, Ptr{Int32}, Ptr{Int32}, Ptr{Float64}, Ptr{Float64}, Ptr{Float64}),
-                 $(io_err),  $(nvar),    $(ncon),    $(x),         $(f),         $(c));
+                 $(io_err),  &$(nvar),   &$(ncon),   $(x),         $(f),         $(c));
```
dpo: Do we want this? Those variables have been set by the call to `cutest_csetup`. I know `cutest_cfn` can also return those values, but I've never liked the fact that you can get them from several places.

abelsiqueira: Actually I only did this because they should be pointers, and without this I get deprecation warnings.
I followed what was used in the other ccalls inside `julia_interface.jl`.

dpo:
Closing since this is included in other PRs.

Issue #22, opened 2015-05-15
Interface with MathProgBase


dpo:
We should do it for compatibility. The [AMPL interface](https://github.com/mlubin/AMPLMathProgInterface.jl) can serve as a model. Seems pretty straightforward.

dpo:
This issue should probably be migrated to NLPModels.jl.

PR #25, opened 2015-06-19
Interface with MathProgBase

abelsiqueira:
Implements #22.

This doesn't have a travis test yet.
I [tested](https://gist.github.com/abelsiqueira/d0a8fa2507b271a77be5) with `Ipopt.jl` in v0.4.

This has room to improvement, especially with the inplace function calls, but I'd rather wait for the intermediate interface, since it will be much cleaner then.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 Compat
+MathProgBase
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -51,6 +51,7 @@ end
 
 include("core_interface.jl")
 include("julia_interface.jl")
+include("MathProgBase_interface.jl")
 
 # Decode problem and build shared library.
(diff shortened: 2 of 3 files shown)


dpo:
Also wanted to draw your attention to https://github.com/dpo/NLPModels.jl

abelsiqueira:
Dropping this in favor of NLPModels.
Also, see [this](https://drive.google.com/file/d/0B93L4n2eB78POUdjZUtNd3BnYlk/view?usp=sharing).

dpo:
That's great, thanks. Was the graph generated automatically? Were you looking at the `develop` branch? There's no type named `NLPModel`, but there's a `JuMPNLPModel`.

abelsiqueira:
I used [draw.io](https://www.draw.io/). I invited you to the graph.

I wasn't looking at `develop` in fact. I'll check it out again.

PR #35, opened 2015-08-04
Require is deprecated on 0.4

abelsiqueira:
`require` is deprecated on 0.4

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -2,7 +2,7 @@
 
 module CUTEst
 
-require(Pkg.dir("MathProgBase","src","NLP","NLP.jl"))
+push!(LOAD_PATH, Pkg.dir("MathProgBase","src","NLP"))
 using NLP  # Defines NLPModelMeta.
 using Compat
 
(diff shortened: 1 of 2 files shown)


src/CUTEst.jl
```diff
@@ -2,7 +2,7 @@
 
 module CUTEst
 
-require(Pkg.dir("MathProgBase","src","NLP","NLP.jl"))
+include(Pkg.dir("MathProgBase","src","NLP","NLP.jl"))
```
dpo: Doesn't seem to work in 0.3:
```JULIA
julia> using CUTEst
ERROR: NLP not found
```

abelsiqueira:
Ah, I still have the old NLP so it didn't break. Travis was also downloading it. I will try to change this another way.

dpo:
Good point. I forgot to change `.travis.yml`!

abelsiqueira:
I hope I've fixed it by using `push!(LOAD_PATH,...)`. This allows `using NLP` without warnings.
Also removed NLP from travis.

dpo:
Works for me. Thanks.

PR #37, opened 2015-08-04
Remove troublesome spaces on documentation

abelsiqueira:
The documentation is converted to Markdown, ignore the line length but not the leading spaces. This could cause multiple spaces between words.

### DIFF
--- a/src/julia_interface.jl
+++ b/src/julia_interface.jl
@@ -2,11 +2,11 @@ export objcons, objgrad, obj, cons_coord, cons, hess_coord, hess, hprod, hprod!
 
 """objcons(nlp, x)
 
-  Computes the objective function and constraint vector values at x.
-  Usage:
+Computes the objective function and constraint vector values at x.
+Usage:
 
-      f, c = objcons(nlp, x) # If the problem is constrained
-      f = objcons(nlp, x)    # If the problem is unconstrained
+    f, c = objcons(nlp, x) # If the problem is constrained
+    f = objcons(nlp, x)    # If the problem is unconstrained
 
   - nlp: [IN] CUTEstModel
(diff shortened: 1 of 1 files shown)


src/julia_interface.jl
```diff
@@ -2,11 +2,11 @@ export objcons, objgrad, obj, cons_coord, cons, hess_coord, hess, hprod, hprod!
 
 """objcons(nlp, x)
```
dpo: Should this be indended 4 spaces as well? It looks a little weird, but then it's formatted as code (it renders in color in my terminal and I presume it's even more fancy in IJulia).

src/julia_interface.jl
```diff
@@ -245,14 +245,13 @@ end
...
+If the problem is constrained, the Hessian is of the Lagrangian function at
+x with Lagrange multipliers y, otherwise the Hessian is of the objective
+function at x.
+Usage:
 
-      Hv = hprod(nlp, x, y, v)
+    Hv = hprod(nlp, x, y, v)
 
   - nlp: [IN] CUTEstModel
   - x:   [IN] Array{Float64, 1}
```
dpo: The documentation for `v` is missing here (check also the other `hprod`s)

dpo:
Beautiful! Many thanks!

src/julia_interface.jl
abelsiqueira: This does not render in color in my terminal, I'll add the 4 spaces.

abelsiqueira: Alternatively, we could use `# objcons(nlp, x)`, which is useful for duplicated functions.

abelsiqueira:
Fixed

PR #40, opened 2015-08-07
Implement grad, grad! and cons!

dpo:


src/julia_interface.jl
```diff
@@ -68,8 +70,61 @@ function objgrad(nlp :: CUTEstModel, x :: Array{Float64,1}, grad :: Bool)
...
+  - x:    [IN] Array{Float64, 1}
+  - g:    [INOUT] Array{Float64, 1}
+"""
+function grad!(nlp :: CUTEstModel, x :: Array{Float64,1}, g :: Array{Float64,1})
+  nvar = nlp.meta.nvar;
+  ncon = nlp.meta.ncon;
+  f = Cdouble[0];
+  io_err = Cint[0];
+  get_grad = 1;
+  cutest_fcn = ncon > 0 ? "cutest_cofg_" : "cutest_uofg_";  # How do you do this with symbols?
```
abelsiqueira: This should be `cutest_ofg`

dpo: You mean the variable name?

src/julia_interface.jl
```diff
@@ -294,7 +414,7 @@ Usage:
   - x:   [IN] Array{Float64, 1}
   - y:   [IN] Array{Float64, 1}
   - v:   [IN] Array{Float64, 1}
-  - Hv:  [OUT] Array{Float64, 1}
+  - Hv:  [INOUT] Array{Float64, 1}
```
abelsiqueira: I'm not very fond of `INOUT` because it seems to me that the value of `Hv` on input would be used. I believe that `OUT` is sufficient to tell the user that this argument is going to be modified. But it's your choice.

src/julia_interface.jl
abelsiqueira: Yes, I actually mistook it for fn (from `ufn` and `cfn`), but I understood now that it means function.

abelsiqueira:
Do you think we should use specialized or core functions instead of the `ccall`s? Since these functions use nlp, if would be very easy.

src/julia_interface.jl
dpo: I see. I made the same mistake and changed it.

src/julia_interface.jl
dpo: I'm happy to remove it. The CUTEst man pages just say `OUT`, so I'll revert that commit.

dpo:
On one hand, it would make the Julia interface quite a bit cleaner. On the other hand, it would add another layer during calls. The slow down would probably not be noticeable though. For example, `objcons` would look something like this:

```JULIA
function objcons(nlp :: CUTEstModel, x :: Array{Float64,1})
  return nlp.meta.ncon > 0 ? ufn(nlp, x) : cfn(nlp.x)
end
```

The call chain would be
* `objcons(nlp, x)` in the Julia interface
* `ufn(nlp, x)` in the specialized interface
* `ufn(io_err, n, x, f, libname)` in the core interface
* `cutest_ufn()` in `libcutest`.

Right now, `objcons` calls `cutest_ufn` directly. But the intermediate functions are just passing scalars and pointers around, there's no copy of arrays.

abelsiqueira:
This might be worth evaluating. The creation of other "end-user" interfaces would be easier, but then, we don't know yet if they'll be required. I'll create an issue for future reference.

src/julia_interface.jl
```diff
@@ -144,6 +199,71 @@ Usage:
 """
 cons(nlp :: CUTEstModel, x :: Array{Float64,1}) = cons_coord(nlp, x, false);
 
+"""    cons!(nlp, x)
```
abelsiqueira: This should be `cons!(nlp, x, c)`

src/julia_interface.jl
```diff
@@ -144,6 +199,71 @@ Usage:
...
+
+Computes the constraint vector value in place.
+Usage:
+
+    c = cons!(nlp, x)
```
abelsiqueira: This should be `cons!(nlp, x, c)`

src/julia_interface.jl
```diff
@@ -144,6 +199,71 @@ Usage:
...
+
+  - nlp: [IN] CUTEstModel
+  - x:   [IN] Array{Float64, 1}
+  - c:   [INOUT] Array{Float64, 1}
```
abelsiqueira: `INOUT` here.

src/julia_interface.jl
```diff
@@ -68,8 +70,61 @@ function objgrad(nlp :: CUTEstModel, x :: Array{Float64,1}, grad :: Bool)
...
+"""    grad!(nlp, x, g)
+
+Computes the objective function gradient at x in place.
+Usage:
+
+    grad!(nlp, x, g)
+
+  - nlp:  [IN] CUTEstModel
+  - x:    [IN] Array{Float64, 1}
+  - g:    [INOUT] Array{Float64, 1}
```
abelsiqueira: `INOUT` here

dpo:
Thanks for all the comments! You've got to love `git rebase`.

abelsiqueira:
> You've got to love `git rebase`.
Don't I know it :)

abelsiqueira:
Thanks for the commit

PR #44, opened 2015-09-01
Fix problem with nnet

abelsiqueira:
Travis broke, but my computer didn't.


dpo:
That's due to a recent change in NLP. `nnet` is now the list of *nonlinear network* constraints. It used to be the *number of network` constraints. There is also `lnet`, which is the list of `linear network` constraints, and `nlnet`, the number of such constraints. Sorry I forgot to make the change here.

PR #46, opened 2015-11-06
Remove `@eval` from everywhere

### LINKED ISSUE #45: Segmentation fault
@dpo, I had a weird segmentation fault (random occurrence), could you try to recreate?
````julia
using CUTEst
nlp = CUTEstModel("ROSENBR")
for i = 1:100000
    uhprod(nlp, false, [0.0;0.0], [0.0;0.0])
end
````

### PR DESCRIPTION
**Pull request by abelsiqueira**
Closes #45

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -13,8 +13,7 @@ cutest_instances = 0;
 
 type CUTEstModel
   meta    :: NLPModelMeta;
-  libname :: ASCIIString;
-  cutest_lib :: Any # Or Ptr{Void} ?
+  cutest_lib :: Ptr{Void}
 end
 
 const cutest_arch  = get(ENV, "MYARCH", "");
@@ -50,6 +49,8 @@ macro cutest_error()  # Handle nonzero exit codes.
   :(io_err[1] > 0 && throw(CUTEstException(io_err[1])))
 end
 
+# Taken from
+# http://docs.julialang.org/en/release-0.4/manual/calling-c-and-fortran-code/#indirect-calls
 macro dlsym(func, lib)
   z, zlocal = gensym(string(func)), gensym()
   eval(current_module(), :(global $z = C_NULL))
@@ -81,7 +82,7 @@ function sifdecoder(name :: ASCIIString)
   run(`rm ELFUN.f EXTER.f GROUP.f RANGE.f ELFUN.o EXTER.o GROUP.o RANGE.o`);
   push!(Libdl.DL_LOAD_PATH,".")
   cutest_lib = Libdl.dlopen(libname)
-  return libname, cutest_lib
+  return cutest_lib
 end
 
 # Initialize problem.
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/CUTEst.jl [hunk 1]
```diff
@@ -13,7 +13,7 @@ cutest_instances = 0;
 
 type CUTEstModel
   meta    :: NLPModelMeta;
-  libname :: ASCIIString;
+  cutest_lib :: Any # Or Ptr{Void} ?
```
dpo: Doesn't `dlopen` return a `Ptr{Void}`?

src/CUTEst.jl
```diff
@@ -49,6 +49,21 @@ macro cutest_error()  # Handle nonzero exit codes.
   :(io_err[1] > 0 && throw(CUTEstException(io_err[1])))
 end
 
+macro dlsym(func, lib)
```
dpo: Better indicate where you got this from, so we remember.

src/CUTEst.jl
```diff
@@ -64,7 +79,8 @@ function sifdecoder(name :: ASCIIString)
   run(`$linker $sh_flags -o $libname.$soname ELFUN.o EXTER.o GROUP.o RANGE.o -L$cutest_dir/objects/$cutest_arch/double -lcutest_double`);
   run(`rm ELFUN.f EXTER.f GROUP.f RANGE.f ELFUN.o EXTER.o GROUP.o RANGE.o`);
   push!(Libdl.DL_LOAD_PATH,".")
-  return libname
+  cutest_lib = Libdl.dlopen(libname)
+  return cutest_lib
```
dpo: I like it!

src/core_interface.jl
```diff
@@ -8,11 +8,11 @@ export usetup, csetup, udimen, udimsh, udimse, uvartype, unames,
 
 function usetup(io_err::Array{Cint, 1}, input::Array{Cint, 1}, out::Array{Cint, 1},
     io_buffer::Array{Cint, 1}, n::Array{Cint, 1}, x::Array{Cdouble, 1},
-    x_l::Array{Cdouble, 1}, x_u::Array{Cdouble, 1}, libname)
-  @eval ccall(("cutest_usetup_", $(libname)), Void,
+    x_l::Array{Cdouble, 1}, x_u::Array{Cdouble, 1}, cutest_lib)
```
dpo: This seems a bit more tricky because users must now pass a pointer! I wonder if this is a recipe for hard crashes. Perhaps these functions should keep taking `libname` and call `dlopen` internally?!

src/julia_interface.jl
```diff
@@ -22,13 +22,13 @@ function objcons(nlp :: CUTEstModel, x :: Array{Float64,1})
   f = Cdouble[0];
   c = Array(Float64, ncon);
   if ncon > 0
-    @eval ccall((:cutest_cfn_, $(nlp.libname)), Void,
+    ccall(@dlsym(:cutest_cfn_, nlp.cutest_lib), Void,
```
dpo: Here it's ok because the pointer is part of the `nlp` object.

src/core_interface.jl
abelsiqueira: Then maybe leave both, first because I like to use the core interface inside `CUTEst.jl` and the specialized interface, and because the time it takes is about 7 times longer with `dlopen`.

dpo: Oh I see. Ok, let's leave it then.

abelsiqueira:
Updated regarding the comments

dpo:
Thanks a lot for investigating this!!!

PR #48, opened 2016-01-05
Fix issue4

### LINKED ISSUE #4: Calling two different problems, one after the other, breaks the second.
At least as far as objective value in concerned.
In https://travis-ci.org/abelsiqueira/CUTEst.jl/jobs/50675367, first we decode HS35 and print the objective value correctly. Then, we decode HS53, and obtain a incorrect value for the objective value.
Things worth noting:
  - If I change the order of tests called in `runtests.jl` (first and second line), then HS53 is correct and HS35 is not.
  - Using the libname as the problem name (the default way of calling `CUTEstModel`), doesn't change anything.
  - If I change HS53 to HS31, it appears to not have a problem. Since HS35 and HS31 have the same dimensions, this appears to be related to memory allocation.
  - As above, if I change HS35 to HS51 or HS52, is appears to work, and again they have the same dimensions.
  - Same for HS32 and HS33.

This leads me to think that this is a problem with CUTEst in Fortran. It appears to use something in the allocation process that it keeps in the memory even after terminate is called. Unfortunately, the code is beyond me.

abelsiqueira:
This allows opening one problem **after the other** without breaking.
Unfortunately, this prohibits the possibility of opening two problems **at the same time**.

Closes #4.

The most important changes are in `src/CUTEst.jl`. The libname or cutest_lib arguments are removed in favor of using a global variable.
Tested running **both** HS32 and HS4 verifying the values and by getting the objective function of **all** the CUTEst problems with the three interfaces (see `test/runtests.jl`).

The fix for travis (which allows failures of osx) is in a different commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ install:
 
 script:
   - julia --check-bounds=yes -E 'Pkg.test("CUTEst"; coverage=true)'
-  - julia --check-bounds=yes -E 'cd(Pkg.dir("CUTEst")); include("test/runtests_unc.jl")'
 
 after_success:
   - julia -e 'cd(Pkg.dir("CUTEst")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ os:
   - linux
   - osx
 
+matrix:
+  allow_failures:
+    - os: osx
+
 julia:
   - release
   - nightly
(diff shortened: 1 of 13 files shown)


abelsiqueira:
Closed in favor of #49.

abelsiqueira:
Updated. Working on `deps/build.jl` now.

.travis.yml
```diff
@@ -4,6 +4,10 @@ os:
   - linux
   - osx
 
+matrix:
+  allow_failures:
+    - os: osx
+
```
dpo: Why is this allowed to fail now?

abelsiqueira: Sorry, missed it.

abelsiqueira: Fixed

PR #49, opened 2016-02-12
Major change on usage. Global NLP

### LINKED ISSUE #4: Calling two different problems, one after the other, breaks the second.

abelsiqueira:
Closes #4.

This commit was built on top of PR #48. The fix in that PR used a global
variable to store the dynamic library and thus made libname obsolete.
Furthermore, `CUTEstModel` collapses into `NLPModelMeta`, and since
we'll have only one open probem at a time, we can use a global variable for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ install:
 
 script:
   - julia --check-bounds=yes -E 'Pkg.test("CUTEst"; coverage=true)'
-  - julia --check-bounds=yes -E 'cd(Pkg.dir("CUTEst")); include("test/runtests_unc.jl")'
 
 after_success:
   - julia -e 'cd(Pkg.dir("CUTEst")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ os:
   - linux
   - osx
 
+matrix:
+  allow_failures:
+    - os: osx
+
 julia:
   - release
   - nightly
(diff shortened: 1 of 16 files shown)


dpo:
Sorry for taking so long to reply. It's great that you fixed the issue as much as possible. I think I preferred the way the problem was defined before, i.e., `CUTEstModel` instead of just `NLPModel` for a couple of reasons:
1. the name suggests the provenance
2. let's not give up hope that we'll fix the problem completely one day and have one library per instance.
So in that sense, I liked #48 better.

But it looks like this PR fixes other things as well (e.g., Travis).

Do you think it's possible to get the best of both PRs?

In all cases, many thanks for your hard work!!!

abelsiqueira:
Hey, no problem. I'll use the other one and update the branch.

Issue #50, opened 2016-02-12
Use deps/build.jl


abelsiqueira:
Suggested by @timholy.

dpo:
We have releases now:
* https://github.com/optimizers/archdefs-mirror
* https://github.com/optimizers/sifdecode-mirror
* https://github.com/optimizers/cutest-mirror
* https://github.com/optimizers/mastsif-mirror
* https://github.com/optimizers/maros-meszaros-mirror

See also https://github.com/optimizers/homebrew-cutest/commits/release

abelsiqueira:
I implemented this on my branch [feat/bindeps](https://github.com/abelsiqueira/CUTEst.jl/tree/feat/bindeps) on top of #48. I'll wait until #48 is merged to prevent a huge PR. Also, it currently does not work on OSX, but I can't test easily.
Despite the branch name, it does not use BinDeps. I tried to use it, but for a non-straightforward build it isn't very helpful. I ended up using Homebrew for OSX and my new package [linux-cutest](https://github.com/abelsiqueira/linux-cutest) for Linux.

The most interesting part is the nonnecessity of setting anything on bashrc, like is usually required.

PR #52, opened 2016-03-27
Create deps/build.jl

abelsiqueira:
`Pkg.build("CUTEst")` downloads and install CUTEst on linux and osx.
Basically I created a installer for Linux, to decrease all the overhead of installing linuxbrew, but for osx I'm still using Homebrew.
Furthermore, to prevent `. $f/$f.bashrc`, I create a `cutestenv.jl` which defines the cutest variables from the installation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+*.[df]
+*.so
+
+deps/*
+!deps/build.jl
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,21 +11,10 @@ julia:
 sudo: required
 
 before_install:
-  # Install Linuxbrew.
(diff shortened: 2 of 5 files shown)


abelsiqueira:
Closes #50

dpo:
I like the idea of `cutestenv.jl`. For the installation, is there no way to download and install each package, essentially as I do in the Homebrew formulas?

abelsiqueira:
Do you mean using linuxbrew?

dpo:
I mean using unix commands as in https://github.com/optimizers/homebrew-cutest/blob/master/sifdecode.rb#L48

abelsiqueira:
I created the linux-cutest package for that. It's the equivalent.

abelsiqueira:
Elaborating, I created [linux-cutest](https://github.com/abelsiqueira/linux-cutest) for the linux installation.
I tried to follow homebrew's formula for the most part.

PR #56, opened 2016-06-21
Bindeps and AbstractNLPModel

abelsiqueira:
Two major changes.

- bindeps: Install CUTEst locally automatically for use inside CUTEst.jl
- use NLPModels.jl: Last "basic" main package for NLPModels.jl (Along with JuMP, Ampl and Simple)

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -20,17 +20,19 @@ cd(dirname(@__FILE__))
 end
 
 @osx_only begin
-  try
-    run(`brew doctor`)
+  @eval using Homebrew
+
+  Homebrew.update()
+  Homebrew.add("homebrew/versions/gcc5")
+  ENV["CUTEST_GCC"] = "gcc-5"
+  ENV["CUTEST_GFORTRAN"] = "gfortran-5"
+  for p in ["archdefs", "cutest", "mastsif"]
+    Homebrew.add("optimizers/cutest/$p")
   end
-  run(`brew update`)
-  run(`brew tap optimizers/cutest`)
-  run(`brew install cutest`)
-  run(`brew install mastsif`)
 
   open("cutestenv.jl", "w") do cenv
     for p in ["archdefs", "cutest", "sifdecode", "mastsif"]
-      prefix = chomp(readall(`brew --prefix $p`))
+      prefix = Homebrew.prefix(p)
       open("$prefix/$p.bashrc") do f
         for line in readlines(f)
           var = split(split(line, "=")[1])[2]
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,0 +1,45 @@
+cd(dirname(@__FILE__))
+
+@linux_only begin
+  lnxurl = "https://raw.githubusercontent.com/abelsiqueira/linux-cutest/master/install.sh"
+  run(`wget $lnxurl -O install.sh`)
+  run(`bash install.sh`)
+
+  open("cutestenv.jl", "w") do cenv
+    open("cutest_env.bashrc") do f
+      for p in ["PATH", "MANPATH", "LD_LIBRARY_PATH"]
+        write(cenv, "$p = get(ENV, \"$p\", \"\")\n")
+      end
+      for line in readlines(f)
+        var = split(split(line, "=")[1])[2]
+        path = chomp(split(line, "=")[2])
+        write(cenv, "ENV[\"$var\"] = \"$path\"\n")
+      end
+    end
+  end
+end
+
+@osx_only begin
+  try
+    run(`brew doctor`)
+  end
+  run(`brew update`)
+  run(`brew tap optimizers/cutest`)
+  run(`brew install cutest`)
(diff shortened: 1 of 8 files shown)


dpo:
Thanks! I'm expecting an issue on OSX because brew currently ships gcc and gfortran 6.1, which are buggy and can't compile sifdecode.

lrsantos11:
@dpo there is any way out to install CUTEst.jl on OSX without this gfortran bug for sifdecode? @abelsiqueira pull request #52 seems to make this gfortran independent?

dpo:
@lrsantos11 Try this:
```bash
$ brew update  # I just updated archdefs.rb
$ brew tap homebrew/versions  # for gcc-5 and gfortran-5, which are ok
$ brew install gcc5
$ CUTEST_GCC=gcc-5 CUTEST_GFORTRAN=gfortran-5 brew install archdefs
$ brew install cutest
```

dpo:
@abelsiqueira `.travis.yml` should probably do the same.

deps/build.jl
```diff
@@ -0,0 +1,45 @@
...
+end
+
+@osx_only begin
+  try
+    run(`brew doctor`)
+  end
+  run(`brew update`)
+  run(`brew tap optimizers/cutest`)
+  run(`brew install cutest`)
+  run(`brew install mastsif`)
```
dpo: Would this work with Homebrew.jl?

abelsiqueira:
I just pushed a possible change using `Homebrew.jl` with the new commands to a different branch.
Since I don't use osx, we'll have to wait for Travis.

lrsantos11:
Thanks @dpo. It's installed like a charm.

abelsiqueira:
@dpo Now using Homebrew.jl.
Julia nightly for OSX is not working for some unknown reason. It hangs on the tests, which may indicate they are taking too long. But they weren't before.

By the way, this supercedes #52.

deps/build.jl
```diff
@@ -0,0 +1,48 @@
...
+        var = split(split(line, "=")[1])[2]
+        path = chomp(split(line, "=")[2])
+        write(cenv, "ENV[\"$var\"] = \"$path\"\n")
+      end
+    end
+  end
+end
+
+@osx_only begin
+  @eval Pkg.add("Homebrew")
```
dpo: This should be in the `REQUIRE` file.

abelsiqueira: But Linux doesn't require it.

dpo: You can use the `@osx_only` syntax in `REQUIRE`.

abelsiqueira:
Force pushed correction

dpo:
Any idea what this is: https://travis-ci.org/JuliaOptimizers/CUTEst.jl/jobs/140419572#L815 ?

abelsiqueira:
I think there is some random error, maybe because of the force push. My homebrew branch had the same error, and I also force pushed it. But my abstractnlp branch was not force pushed and did not produce error.
Notice the [commits](https://travis-ci.org/abelsiqueira/CUTEst.jl/builds).

I'm gonna ask for a restart in the build. Let's see what happens.

abelsiqueira:
Magically works now.

deps/build.jl
```diff
@@ -0,0 +1,47 @@
...
+        var = split(split(line, "=")[1])[2]
+        path = chomp(split(line, "=")[2])
+        write(cenv, "ENV[\"$var\"] = \"$path\"\n")
+      end
+    end
+  end
+end
+
+@osx_only begin
+  @eval using Homebrew
```
dpo: I don't think there's a need to `@eval` here.

deps/build.jl
```diff
@@ -0,0 +1,47 @@
...
+
+  Homebrew.update()
+  Homebrew.add("homebrew/versions/gcc5")
```
dpo: Need to `tap homebrew/versions`.

dpo: Ah, maybe not.

deps/build.jl
abelsiqueira: @eval was due to some error regarding `using` not being on the top level.
There was some explaining that I didn't quite understood, but using @eval
fixed the problem.
Some simple code that may reproduce the error is
````
@osx_only begin
using Plots
for i = 1:3
    println(i)
end
end
````

dpo: Strange. I do it in `HSL.jl`.

dpo:
Actually, if the user already has CUTEst installed, this reinstalls the entire thing, including gcc-5. It'd be better to detect if `libcutest` is already available or not. See what I did in MUMPS.jl for example.

abelsiqueira:
I'm gonna create a issue to tackle this later.

dpo:
@abelsiqueira @lrsantos11 Homebrew now ships gcc/gfortran 6.2, which don't have the bug above. So there's no need to use gcc-5 anymore.

PR #57, opened 2016-06-27
Only return lower triangle of Hessian

dpo:


dpo:
This is a sneaky source of bugs for us. AMPL and JuMP models only return the lower triangle.

@abelsiqueira How should the specialized/core tests be updated? In your code-generation scripts? Could you please indicate how?

abelsiqueira:
I think it's easier to not change `hs32.jl` nor the specialized/core tests, and change only the julia tests.

dpo:
Done, thanks. I think the test that goes through all SIF files is too extreme.

abelsiqueira:
This is not based on the latest develop. I changed to use only 5% of the sifs.

dpo:
Sorry. Here we go.

abelsiqueira:
Thank you

PR #63, opened 2016-06-29
Changes for consistency on NLPModels

abelsiqueira:
Change 1e20 to Inf and -1e20 to -Inf.
Change hess related functions to use keyword y.

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -126,6 +126,11 @@ function CUTEstModel(name :: ASCIIString; decode :: Bool=true, verbose = false)
   end
   @cutest_error
 
+  for lim in Any[bl, bu, cl, cu]
+    I = find(abs(lim) .>= 1e20)
+    lim[I] = Inf * lim[I]
+  end
+
   lin = find(linear);
   nln = setdiff(1:ncon, lin);
   nlin = sum(linear);
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -8,6 +8,8 @@ for problem in ["HS32", "HS4"]
   include("finalize_test.jl")
 end
 
+include("test_mpb.jl")
+
 problems = randsubseq(readdir(get(ENV, "MASTSIF", "") ), 0.05)
 
 for p in problems
@@ -44,5 +46,3 @@ for p in problems
   end
   cutest_finalize(nlp)
 end
-
-include("test_mpb.jl")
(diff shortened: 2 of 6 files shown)


test/runtests.jl
```diff
@@ -44,5 +46,3 @@ for p in problems
   end
   cutest_finalize(nlp)
 end
-
-include("test_mpb.jl")
```
dpo: I liked it here because it's a nice-looking conclusion to the tests ;-).

test/test_mpb.jl
```diff
@@ -10,3 +10,4 @@ model = NLPtoMPB(nlp, IpoptSolver())
 @assert isa(model, Ipopt.IpoptMathProgModel)
 MathProgBase.optimize!(model)
 @assert MathProgBase.getobjval(model) ≈ 0.0
+cutest_finalize(nlp)
```
dpo: Good catch!

dpo:
Looks good!

dpo:
I noticed this earlier: https://travis-ci.org/JuliaOptimizers/CUTEst.jl/jobs/141084034#L3870
Only on Linux and seems intermittent.

abelsiqueira:
I think it may be related to v0.5. The error is in a matrix product and the dimensions should match. This may be related to global scope, or reloading the definitions.
Maybe we should use allow_failure for nightly?

dpo:
Is it a bug we can report? If not, I guess we can use `allow_failure`.

abelsiqueira:
Would have to investigate further and try to create a MFE. Don't have much time for that now.

dpo:
It's intermittent. Now the test passed.

src/CUTEst.jl
```diff
@@ -126,6 +126,11 @@ function CUTEstModel(name :: ASCIIString; decode :: Bool=true, verbose = false)
   end
   @cutest_error
 
+  for lim in Any[bl, bu, cl, cu]
+    I = find(abs(lim) .>= 1e20)
+    lim[I] = Inf * lim[I]
+  end
+
```
dpo: I'm not so sure how CUTEst subroutines will handle those. What will Julia pass Fortran in place of those `Inf`s?

abelsiqueira:
> I noticed this earlier: https://travis-ci.org/JuliaOptimizers/CUTEst.jl/jobs/141084034#L3870
> Only on Linux and seems intermittent.

It happened on the OSX too:
https://travis-ci.org/abelsiqueira/CUTEst.jl/jobs/141866002#L868

src/CUTEst.jl
abelsiqueira: From my tests, Fortran gets an `Infinity`. It's a little confusing to understand if Infinity is always present on Fortran, but it looks like it is on gfortran.
To compare
````Fortran
real*4 x, y
x = 1.0
y = 0.0
write (*,*) x/y
````
returns Infinity.

abelsiqueira:
Rebased

PR #67, opened 2016-07-03
update to README.md

mcprentiss:
This is a small edit, which I think explains this work better and makes it more searchable.


abelsiqueira:
Thanks for the suggestion. There is an open pull request modifying README.md, so I included something similar there.

PR #69, opened 2016-09-03
Pass arguments to SIF decoder

dpo:


test/runtests.jl
```diff
@@ -46,3 +46,12 @@ for p in problems
   end
   cutest_finalize(nlp)
 end
+
+# test arguments passed to decoder
+model = CUTEstModel("DIXMAANJ", "-param", "M=5")
```
abelsiqueira: These should be `nlp` or L53 and L57 should be `model`.

dpo: Gee, thanks! :-}

abelsiqueira:
Thanks

pkofod:
Awesome!

dpo:
That gives access to A LOT more problems!

PR #73, opened 2016-09-09
Select problems

abelsiqueira:
This implements a classification of the CUTEst problems using `$MASTSIF/CLASSF.DB` and the problem itself opened in Julia.

This also implements a select tool based on that classification. It should be useful along with JuliaSmoothOptimizers/Optimize.jl

This also uncovered some problems with the MASTSIF classification, which can be seen at the travis logs.


src/CUTEst.jl
```diff
@@ -104,7 +105,7 @@ function CUTEstModel(name :: ASCIIString, args...; decode :: Bool=true, verbose
     cutest_lib = Libdl.dlopen(libname,
         Libdl.RTLD_NOW | Libdl.RTLD_DEEPBIND | Libdl.RTLD_GLOBAL)
   else
-    sifdecoder(name, args..., verbose=verbose)
+    sifdecoder(ascii(name), args..., verbose=verbose)
```
dpo: `sifdecoder` should probably also take an `AbstractString` as input?

dpo:
This is a lot of work! How do you envision problem selection should go for other problem sources (AMPL, JuMP, Simple)?

src/CUTEst.jl
abelsiqueira: The alternative to ASCIIString is UTFxString, and since I don't know how the command line sifdecoder will handle that, I made that change. If you think the alternative is better, I'll make the change

abelsiqueira:
I think the only other selection we need to worry is the JuliaSmoothOptimizers/OptimizationProblems.jl#3. CUTEst is a special case among the NLPModels, because it has the repository as main objective.

This branch shows a possible course of action, for OptimizationProblems. We define static characteristics that are not easy to verify (like the MASTSIF table), and query other characteristics creating a .json file (or a database?) and update the json every time there is a new problem. The selection function opens the json file, which is relatively fast.
Advantages (change json to another DB, if desired)
- Once the json is created, querying is relatively fast.
- The json can be used by other softwares/packages (online selector?)
- If we define a json standard, a simgle NLPQuery can be created.

Disavantages
- Every new problem needs a manual update.
- Demands additional static characteristics.

dpo:
Right. I like the JSON table, the flexibility it offers and how it's a lot more future-proof than the `CLASSF.DB` file. I'm thinking a SIF-style classification string could be supplemented with data we gather by instantiating problems. The `meta` data structure contains a lot of info. For example, `CLASSF.DB` currently doesn't let you select equality-constrained problems only or inequality-constrained problems only.

src/CUTEst.jl
dpo: What I'm saying is that the Julia function `sifdecoder()` should probably accept an `AbstractString`. I presume string interpolation does the right thing.

abelsiqueira:
Updated

dpo:
I'm quite a bit behind reviewing this. @vepiteski It would be great if you could also test this.

dpo:
NB: There's a new `ready to go` label that you can apply to PRs to indicate that you're finished with the changes.

abelsiqueira:
Rebased

timholy:
FYI this is working very well for me.

dpo:
Thanks. I'm swamped and can't find a chance to test. I propose we merge this so others can start using it. Many thanks @abelsiqueira !

PR #76, opened 2016-09-20
Docs for specialized and core interfaces and other corrections

abelsiqueira:
On the specialized/core front, I changed the scripts to add the documentation right before the functions.
This includes the simple usage of each and a header from the man page on the core functions.
This lead to a unfixed problem, so far, that in the REPL, the docs are not in the correct order. Couldn't fix it.
The file `documentation.jl` is gone.

Then I change the `julia_interface.jl` to use `@doc (@doc NLPModels.foo)`. I made some changes on the pages, contents and index too. Shouldn't change much.


dpo:
Looks fantastic, thanks a lot for this hard work!

src/core_interface.jl
```diff
@@ -6,6 +6,32 @@ export usetup, csetup, udimen, udimsh, udimse, uvartype, unames,
...
+"""# usetup
+The usetup subroutine sets up the correct data structures for
+subsequent computations in the case where the only possible
+constraints are bound constraints. The problem under consideration is
+to minimize or maximize an objective function f(x) over all x ∈ Rn
+subject to the simple bounds xl≤x≤xu. The objective function is group-
+partially separable.
+
+This help was generated automatically and may contain
+errors. For more information, run the shell command
```
dpo: I'm not sure we should include this. What kind of errors could there be? In any case, it's good to refer to the man page.

abelsiqueira: Script might have some unverified error. I'm gonna remove the warning.

abelsiqueira:
I removed the script warning, and while doing that discovered new CUTEst functions (changed computer).

dpo:
That's a lot of new stuff!!!

PR #77, opened 2016-09-21
Fix hess_coord when obj_weight is zero

abelsiqueira:
Found with `dercheck.jl`.

### DIFF
--- a/src/julia_interface.jl
+++ b/src/julia_interface.jl
@@ -416,7 +416,7 @@ function hess_coord(nlp :: CUTEstModel, x :: Array{Float64,1}; y :: Array{Float6
                  io_err,     &nvar,      &ncon,      x,            y,            this_nnzh,  &nnzh,      hval,         hrow,       hcol);
     @cutest_error
     nlp.counters.neval_hess += 1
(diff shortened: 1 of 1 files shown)


dpo:
We should actively use `dercheck` in the tests.

abelsiqueira:
I'm putting it on `dercheck`, it shouldn't be a problem anymore.
I'm still having problems with `AmplNLReader` though.

dpo:
Do you mean that you get erroneous derivatives?

abelsiqueira:
Yes. In this case, the Hessian. I'm not sure how Ampl works, but I'm calling `hess(nlp_ampl, nlp_ampl.meta.x0)` right after `consistency_functions` and the value is not what's expected.

dpo:
My guess is that it's the usual AMPL trap where you have to evaluate the function at x before you can evaluate its Hessian at x. Try calling `obj()` at the desired point before calling `hess()`.

abelsiqueira:
Yes, that was it. Didn't know about that. It's fixed now. This can me merged.

dpo:
Thanks!

Issue #80, opened 2016-09-24
Warnings


dpo:
I'm getting the warnings
```
WARNING: Method definition cchprods!(Int64, Int64, Bool, Array{Float64, 1}, Array{Float64, 1}, Int64, Array{Float64, 1}, Array{Int64, 1}, Array{Int64, 1}) in module CUTEst at /Users/dpo/.julia/v0.5/CUTEst/src/specialized_interface.jl:5294 overwritten at /Users/dpo/.julia/v0.5/CUTEst/src/specialized_interface.jl:5304.
WARNING: Method definition cchprods!(CUTEst.CUTEstModel, Bool, Array{Float64, 1}, Array{Float64, 1}, Int64, Array{Float64, 1}, Array{Int64, 1}, Array{Int64, 1}) in module CUTEst at /Users/dpo/.julia/v0.5/CUTEst/src/specialized_interface.jl:5351 overwritten at /Users/dpo/.julia/v0.5/CUTEst/src/specialized_interface.jl:5363.
```

Also
```
WARNING: Method definition f(Any) in module Main at /Users/dpo/.julia/v0.5/CUTEst/test/hs32.jl:4 overwritten at /Users/dpo/.julia/v0.5/CUTEst/test/hs4.jl:4.
WARNING: Method definition g(Any) in module Main at /Users/dpo/.julia/v0.5/CUTEst/test/hs32.jl:5 overwritten at /Users/dpo/.julia/v0.5/CUTEst/test/hs4.jl:5.
WARNING: Method definition H(Any) in module Main at /Users/dpo/.julia/v0.5/CUTEst/test/hs32.jl:8 overwritten at /Users/dpo/.julia/v0.5/CUTEst/test/hs4.jl:6.
WARNING: Method definition #H(Array{Any, 1}, Main.#H, Any) in module Main overwritten.
```
We should probably give each problem function names other than just `f`, `g` and `H`.

abelsiqueira:
I'm looking into the first problem.

`f` and friends are remnants from the days of yore. I'll remake the initial tests to use `NLPModels`.

These will be two PRs after 0.5.

abelsiqueira:
Okay, the problem happens in the script because when
- in an inplace method,
- an output variable
- is an Array{Int, 1}
then I create two methods: one using Array{Int, 1} and one using Array{Cint, 1}.
[...]

PR #82, opened 2016-09-25
Rework interface testing

abelsiqueira:
This uses NLPModels test files to test the interfaces.
Limits scope of variables, preventing redefinition warnings.

Related to #80, the second part. The first part is still not fixed.
Depends on JuliaSmoothOptimizers/NLPModels.jl/pull/47

The diff here isn't handling very well these large changes, so I'll explain a bit.
Before, to test if the interfaces were well formed I created two tests: hs4 and hs32. I would create functions `f`, `g`, etc., build the `CUTEstModel`, test and finalize. Similar do consistency checks.
Now, each interface test is a function receiving a CUTEstModel and another NLPModel to compare.
The tests are pretty much the same, but I changed every `@test_approx_eq_eps` to `@fact --> roughly` because it has built in `rtol`.
On test_{julia,core,specialized} I mainly changed the header and `@test` to `@fact`. Had a couple of fixes on the specialized interface tests.


abelsiqueira:
This was rebased, and together with #90, should close #80.

dpo:
Many thanks!

PR #85, opened 2016-10-06
Attempt to detect a working CUTEst setup

dpo:


dpo:
This works for me locally with an existing installation of CUTEst. Could you also test at your end?

abelsiqueira:
Yes, it is! Thank you.

dpo:
Not sure why it's decided to build gcc from source on osx...

abelsiqueira:
On homebrew-cutest it doesn't?

dpo:
Ah it's because osx 10.9 is no longer supported by Homebrew. I need to request a more recent one.

dpo:
Actually, the problem is something else. When installing CUTEst on 10.11, the gcc bottle is poured: https://travis-ci.org/optimizers/homebrew-cutest/jobs/165556609#L150

But when using Julia Homebrew, it isn't... Not sure what the problem is...

dpo:
I temporarily added a `travis_wait` to see if the build completes.

dpo:
Are osx failures a Travis bug??? I can't see another explanation at the moment.

dpo:
Ok, I think I understand. It's just that the gcc bottles in Homebrew aren't relocatable, so Homebrew.jl compiles from source.

abelsiqueira:
Sorry I haven't been helful. I really don't know much about osx and homebrew.
Can I merge this?

dpo:
No problem. I cleaned up the PR. If the build passes, it should be good to go.

abelsiqueira:
Thank you.

PR #90, opened 2016-10-25
Simplify specialized interface

### LINKED ISSUE #89: Simplify?
CUTEst has too many methods. Besides the core functions and the NLPModels variant, there is also the specialized interface, which I'm not completely sure how useful it is.

For each core function, the specialized interface has between one and six versions of that function.

- One returning all outputs
- One inplace version when there are arrays in the output
- One additional inplace version when there are integer arrays in the output (`Int`->`Cint`)
- If some inputs can be replaced by `nlp`, then another version for each of the previous ones.

See [ugh](https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/develop/src/specialized_interface.jl#L1065), for an example using all six.

Do we need all these versions?

I think a leaner solution would be creating only two interfaces

- Both using `nlp`
- One returning all outputs
- One inplace

abelsiqueira:
Closes #89

### DIFF
--- a/docs/src/core.md
+++ b/docs/src/core.md
@@ -40,15 +40,12 @@ In Julia, we have
 
 ```julia
 f = ufn(n, x)
-f = ufn(nlp, x)
 ```
 
 In other words, a simplification of the original function, returning what is
 simple to return, and reducing the parameters to only what is needed.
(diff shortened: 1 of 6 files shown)


dpo:
Thank you!

Issue #51, opened 2016-02-12
Two problems at the same time


abelsiqueira:
Is it possible to open two problems at the same time?

Something in the lines of

``` julia
nlp1 = CUTEstModel("HS35")
nlp2 = CUTEstModel("BARD")
f1 = obj(nlp1, nlp1.meta.x0)
f2 = obj(nlp2, nlp2.meta.x0)
```

timholy:
Interestingly, even with #97 this doesn't work:
```jl
julia> nlp1 = CUTEstModel("HS35")
Minimization problem HS35
nvar = 3, ncon = 1 (1 linear)


julia> nlp2 = CUTEstModel("BARD")
ERROR: CUTEst.CUTEstException(1,"memory allocation error")
 in #CUTEstModel#13(::Bool, ::Bool, ::Type{T}, ::String) at /home/tim/.julia/v0.5/CUTEst/src/CUTEst.jl:138
 in CUTEst.CUTEstModel(::String) at /home/tim/.julia/v0.5/CUTEst/src/CUTEst.jl:99

julia> finalize(nlp1)

julia> nlp2 = CUTEstModel("BARD")
Minimization problem BARD
nvar = 3, ncon = 0 (0 linear)
```
That line corresponds to the call to `:fortran_open_`, so it's generated by the library rather than julia.

I think that's a strong argument against #97.

abelsiqueira:
Actually, I think the problem is in `usetup`.

It seems the problem is the use of global variables inside fortran `CUTEst`. I don't think there is any interest in changing that, so maybe this should be closed as `wontfix`? @dpo, what do you think?

anriseth:
I was hoping to run lots of CUTEst problems in parallel for some benchmarking. Is that possible currently / are there any hopes of enabling that?

abelsiqueira:
Currently not possible, and maybe impossible without changing the CUTEst source.
I have updated [the branch](https://github.com/abelsiqueira/CUTEst.jl/tree/th-multi) with non-global `cutest_lib` in case you want to try to work this out. The example above fails in the same way in the new branch.

dpo:
Actually, it depends what kind of parallelism you have in mind. If you use a queuing system such as Sungrid Engine or LSF to dispatch problems on different, independent machines, and if a fresh CUTEst module is initialized independently on each machine, you should be ok.

anriseth:
[...]

PR #97, opened 2016-11-10
WIP: eliminate global cutest_lib

timholy:
I was writing code to test different models in a loop, and got the error "CUTEst: call cutest_finalize on current model first". I followed that advice, and encountered a nasty problem:
- call `CUTEst.cutest_finalize(nlp)`
- load a new problem and begin optimization
- at some random point, Julia's gc runs and triggers the finalizer on the original problem. This sets the global variable `cutest_lib` to `NULL` right in the middle of an optimization. Boom.

I didn't think to just try `finalize`, which would have worked much better. Instead, I reworked this package so that
- Each `nlp` object stores its own lib pointer
- low-level calls pass this pointer to wrappers of the library routines

This eliminates the need to manually call `finalize`. As a side benefit, it also lets you have multiple problems loaded at once. It's a fairly big change, so I'll understand if you don't want it.

The WIP is because the documentation needs to be updated, too.

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -9,12 +9,10 @@ using Compat
 import Compat.String
 import Base.Libdl.dlsym
 
-# Only one problem can be interfaced at any given time.
-global cutest_instances = 0
-
 export CUTEstModel, sifdecoder
 
 type CUTEstModel <: AbstractNLPModel
+  lib     :: Ptr{Void}
   meta    :: NLPModelMeta;
 
   counters :: Counters
@@ -45,7 +43,6 @@ type CUTEstException <: Exception
 end
 
 function __init__()
-  global cutest_lib = C_NULL
   deps = joinpath(dirname(@__FILE__), "../deps")
   cutestenv = joinpath(deps, "cutestenv.jl")
   ispath(cutestenv) && include(cutestenv)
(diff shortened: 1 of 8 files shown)


dpo:
Thanks @timholy! The initial version of this package used to do just that: store the lib pointer inside the NLP. We ran into troubles that I can't recall, and it was changed to a global variable. It would be great if we could solve the multiple problem conundrum. Perhaps @abelsiqueira remembers better than I do.

gragusa:
Wherever I go I see @timholy gets there first. I was also looking a CUTESt in a loop and run into the same issue. I support this PR or any other  change that allows this.

timholy:
If it turns out there is some problem with this strategy (e.g., if there's some library that stores global state for the most recently-loaded problem), then a simpler approach will be to change that error message to direct users to call `finalize` rather than `cutest_finalize`.

abelsiqueira:
Hello @timholy . Thanks for this. I didn't realize `gc` could break it, thanks for catching it.
The second solution, direct to `finalize` was the intended use, I did make a mistake.

I don't really remember if there were a main reason to not use `cutest_lib` inside `nlp`. One reason was the additional trouble of passing `cutest_lib` to the specialized interface. However, we now believe that the specialized interface will be mostly ignored in lieu of the `NLPModels` interface.

On the other hand, the global `cutest_instances` is used also because of the "multiple problem conundrum" (#51). Could you test it?

timholy:
Closing due to https://github.com/JuliaSmoothOptimizers/CUTEst.jl/issues/51#issuecomment-259796267

timholy:
I'll leave the branch around just in case someone comes up with a workaround, but feel free to delete it whenever you feel like it.

PR #98, opened 2016-11-10
Change cutest_finalize to finalize on the last wrong places

abelsiqueira:
Despite the discussion on #97 , this is an error, and I'm just fixing it. Questions there still apply, but people using it in the meantime shouldn't be as confused as before.


dpo:
Thanks. Sorry I've been swamped.

PR #99, opened 2016-11-19
select: more flexible canonicalization of objtype/contype

timholy:
Formerly it did not accept ranges, e.g., `objtype=1:4`. This supports any `AbstractVector` type.


timholy:
I've verified that the test failures on 0.4 are unrelated, and fixed by changes similar to #98. (Would be great to merge that so that future PRs get accurate test results.)

timholy:
Rebased on the latest `develop`, and the tests pass.

abelsiqueira:
Thank you

PR #101, opened 2016-11-21
Add select tool to docs

abelsiqueira:


docs/src/tutorial.md
```diff
@@ -132,3 +132,88 @@ print(nlp.meta)
 ```@example ex2
 finalize(nlp)
 ```
+
+## Selection tool
+
+CUTEst comes with a simple selection tool. It uses a static file generated from
+the original `CLASSF.DB` and an execution of each problem.
+Don't try to regenerate the classification file. If you think there is a
```
dpo: I would remove the last two sentences. It's valid to generate a classification file using `classify`. For instance, if a user creates their own collection of problems by pulling from MASTSIF, MarosMeszaros, etc.

docs/src/tutorial.md
```diff
@@ -132,3 +132,88 @@ print(nlp.meta)
...
+```@example exs
+using CUTEst # hide
+problems = CUTEst.select()
+length(problems)
+```
+
+The list of keyword arguments to filter the problems is given below, with their
+default value.
+
+| arguments | default | Description |
```
dpo: `description` shouldn't be capitalized

docs/src/tutorial.md
```diff
@@ -132,3 +132,88 @@ print(nlp.meta)
...
+|-----------|---------|-------------|
+| `min_var` | `0`     | Minimum number of variables |
```
dpo: Shouldn't `min_var` be 1 by default? A problem with no variables doesn't make sense.

docs/src/tutorial.md
```diff
@@ -132,3 +132,88 @@ print(nlp.meta)
...
+| `max_var` | `Inf`   | Maximum number of variables |
+| `min_con` | `0`     | Minimum number of constraints (not bounds) |
+| `max_con` | `Inf`   | Maximum number of constraints |
+| `obj_type` | *all types* | Type of the objective function. Can be one or an array of the following: "none", "constant", "linear", "quadratic", "sum_of_squares", "other", or the corresponding characters: "N", "C", "L", "Q", "S", "O". |
+| `con_type` | *all types* | Type of the constraints. Can be one or an array of the following: "unc", "fixed_vars", "bounds", "newtork", "linear", "quadratic", "general", or the corresponding characters: "U", "X", "B", "N", "L", "Q", "O". |
+| `only_free_var` | `false` | Whether to exclude problems with bounded variables. |
```
dpo: Maybe we should refer to another table that explains the meaning of each `obj_type` and `con_type`?!

docs/src/tutorial.md
```diff
@@ -132,3 +132,88 @@ print(nlp.meta)
...
+| `only_bnd_var` | `false` | Whether to exclude problem with any free variables. |
```
dpo: I would have thought `only_bnd_var` means that the only constraints are bound constraints. But that doesn't mean all variables must be bounded.

docs/src/tutorial.md
```diff
@@ -132,3 +132,88 @@ print(nlp.meta)
...
+| `only_linear_con` | `false` | Whether to exclude problems with nonlinear constraints. |
+| `only_nonlinear_con` | `false` | Whether to exclude problems with any linear constraints. |
```
dpo: Do bounds count as linear constraints?

docs/src/tutorial.md
dpo: I find `only_free_var` a bit confusing. Do you mean no bound constraints? Variables subject to, say, nonlinear constraints can't really be called "free".

docs/src/tutorial.md
```diff
@@ -132,3 +132,88 @@ print(nlp.meta)
...
+
+The selection tool is not as complete as we would like. Some combinations are
+still hard to create. Below we create some of the simpler ones.
+
+### Unconstrained problems
+
+No constraints and no bounds. There are two options:
+
+```@example exs
+problems = CUTEst.select(max_con=0, only_free_var=true)
```
dpo: What's the difference between `only_free_var=true` and `con_type="bound"`?

docs/src/tutorial.md
abelsiqueira: Ah, ok, changing the `CLASSF.DB` file itself. Will remove the lines.

docs/src/tutorial.md
abelsiqueira: Free as [here](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/master/src/nlp_types.jl#L55), i.e., no lower nor upper bounds on the variables.

docs/src/tutorial.md
abelsiqueira: I'm considering variable bounds as in $\ell_i \leq x_i \leq u_i$. The distinction seems important because I want to be able to say equality constraints and bounds on the variables, without excluding each other.

docs/src/tutorial.md
abelsiqueira: Considering `min f(x) s.t cL <= c(x) <= cU, L <= x <= U`, `only_free_var=true` means that `L=-Inf` and `U=Inf`, without considering `c(x)` at all. `contype="bounds"` means that `length(c) = 0` and that `L != -Inf` or `U != Inf`.

docs/src/tutorial.md
abelsiqueira: Just clarifying,
- CUTEst consider a "xLxx" problem a linearly constrained problem. I believe that variables can be unbounded, bounded or fixed.
- The number of linear constraints on the problem is `nlp.meta.nlin`, i.e., I open the problem and store this value.
- `only_linear_con=true` assures that `nlp.meta.nlin` is equal to `nlp.meta.ncon`. It also doesn't look the variable`s bounds situation.
- `only_nonlinear_con=true` assures that `nlp.meta.nlin` is zero. It also doesn't look at the bounds of the variables. In this sense, bounded variables do not count as linear constraints.

abelsiqueira:
I updated with the changes, but the naming can be changed. I'm only trying to create two axis of choice: variables and their bounds and constraint functions type.

dpo:
Is the build error expected here?

abelsiqueira:
No. It's something on Homebrew. [Travis](https://travis-ci.org/JuliaSmoothOptimizers/CUTEst.jl/jobs/177777125#L334).

dpo:
Thank you!

Issue #103, opened 2016-11-22
Store files somewhere beside /tmp?


timholy:
Every time I reboot my laptop (e.g., after security updates that affect the kernel) I need to do `Pkg.rebuild("CUTEst")`. Any thoughts on storing the files elsewhere?

dpo:
What's in /tmp that requires a rebuild?

timholy:
`/tmp` gets wiped when you reboot, and
```sh
$ cat cutestenv.jl 
PATH = get(ENV, "PATH", "")
MANPATH = get(ENV, "MANPATH", "")
LD_LIBRARY_PATH = get(ENV, "LD_LIBRARY_PATH", "")
ENV["ARCHDEFS"] = "/tmp/tmplJ4BFj/archdefs"
ENV["CUTEST"] = "/tmp/tmplJ4BFj/cutest"
ENV["SIFDECODE"] = "/tmp/tmplJ4BFj/sifdecode"
ENV["MASTSIF"] = "/tmp/tmplJ4BFj/mastsif"
ENV["MYARCH"] = "pc64.lnx.gfo"
ENV["PATH"] = "/tmp/tmplJ4BFj/cutest/bin:/tmp/tmplJ4BFj/sifdecode/bin:$PATH"
ENV["MANPATH"] = "/tmp/tmplJ4BFj/cutest/man:/tmp/tmplJ4BFj/sifdecode/man:$MANPATH"
ENV["LD_LIBRARY_PATH"] = "/tmp/tmplJ4BFj/lib:$LD_LIBRARY_PATH"
```

timholy:
Perhaps this is an unintentional consequence of the build script writing its log files to a directory created by `mktempdir`? I naively assumed this was intentional.

abelsiqueira:
Really not intentional. Looks like a result of [this line](https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/develop/deps/build.jl#L11).

timholy:
I'll fix. I'm also curious about [this line](https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/c69f7eeb810d09327b25d08b2f53506b0d328b4a/deps/build.jl#L35). Why delete the file if everything is already installed?

dpo:
> Really not intentional. Looks like a result of this line.

Ah I guess I naively thought `cd` would "come back" when the function exits.

[...]

PR #104, opened 2016-11-22
Build CUTEst inside deps/

timholy:
The first commit downloads the files into a directory `deps/files`. After checking for a "global" installation (ref https://github.com/JuliaSmoothOptimizers/CUTEst.jl/issues/103#issuecomment-262296220), it then checks for a "local" (julia-specific) install. Only if the latter is not found does it download and install the files. EDIT: this will avoid unnecessary re-installations whenever you tag a new version of CUTEst (`Pkg.build` runs every time a package gets updated).

The second commit may be more controversial. It adds a new environment variable, `CUTEst problems` which points to the directory `deps/files/problems`. It leverages this to build all the `*.so` files (and the `AUTOMAT.d` and `OUTSDIF.d` files) inside this folder, rather than the user's current folder. I've been deleting `.so` files from various folders that I run CUTEst problems in, so centralizing these files seems like a nice way to prevent clutter. On the other hand, I wonder if it's currently possible to have two different julia sessions running in two different directories, both working on CUTEst problems; if this gets centralized, that won't be possible anymore (I presume). If so, I guess one option would be to build them in a temporary folder, and have each julia session generate a different random temporary.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -10,14 +10,15 @@ function validate_libcutest()
   tmpdir = mktempdir()
   cd(tmpdir)
   try
-    outlog = tempname()
-    errlog = tempname()
-    run(pipeline(`$runcutest -p genc -D $hs3`, stdout=outlog, stderr=errlog))
-    print(readstring(errlog))
+    cd(tmpdir) do
+      outlog = tempname()
+      errlog = tempname()
+      run(pipeline(`$runcutest -p genc -D $hs3`, stdout=outlog, stderr=errlog))
+      print(readstring(errlog))
+    end
     return true
-  catch
-    return false
   end
+  false
 end
 
 function check_env()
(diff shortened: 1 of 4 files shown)


timholy:
I should add, this fixes #103.

timholy:
The OSX errors arise from installation (click to expand the `julia -E 'Pkg.build("CUTEst");'`), specifically a `Error: no implicit conversion of nil into String` apparently in the Homebrew ruby code. I don't think that's related to this PR, but perhaps someone more knowledgeable than me about OSX should take a peek?

dpo:
Must be a recent change in Homebrew. I'll have a look asap.

dpo:
I am wondering if the issue comes from Travis/OSX because I can't reproduce locally. I "fixed" the Homebrew formula for CUTEst to make Travis happy and now it installs, but there's an error further down, in the tests. I can't reproduce them locally either.

abelsiqueira:
I restarted CUTEst develop branch and it works for 0.4 and 0.5. There is something here that is breaking OSX. Look at [this line](https://travis-ci.org/abelsiqueira/CUTEst.jl/jobs/183338424#L591). It shows that CUTEst on OSX was installed on a tmp dir.

deps/build.jl
```diff
@@ -34,39 +35,58 @@ if validate_libcutest()
...
+          prefix = Homebrew.prefix(p)
+          open("$prefix/$p.bashrc") do f
+            for line in readlines(f)
+              var = split(split(line, "=")[1])[2]
+              path = chomp(split(line, "=")[2])
+              write(cenv, "ENV[\"$var\"] = \"$path\"\n")
+            end
           end
         end
+        println(cenv, "ENV[\"CUTEst problems\"] = \"$(pwd())/problems\"")
```
abelsiqueira: This pwd() is possibly wrong.

deps/build.jl
```diff
@@ -34,39 +35,58 @@ if validate_libcutest()
...
       end
     end
-  end
 
-  @static if is_linux()
-    lnxurl = "https://raw.githubusercontent.com/abelsiqueira/linux-cutest/master/install.sh"
-    run(`wget $lnxurl -O install.sh`)
-    run(`bash install.sh --install-deps`)
+    @static if is_linux()
+      cd(here) do
```
abelsiqueira: This is missing for OSX

deps/build.jl
```diff
@@ -34,39 +35,58 @@ if validate_libcutest()
...
+            open("cutest_env.bashrc") do f
+              for p in ["PATH", "MANPATH", "LD_LIBRARY_PATH"]
+                write(cenv, "$p = get(ENV, \"$p\", \"\")\n")
+              end
+              for line in readlines(f)
+                var = split(split(line, "=")[1])[2]
+                path = chomp(split(line, "=")[2])
+                write(cenv, "ENV[\"$var\"] = \"$path\"\n")
+              end
+              println(cenv, "ENV[\"CUTEst problems\"] = \"$(pwd())/problems\"")
```
dpo: I don't think this is necessary; all problems in that folder are already in `ENV["MASTSIF"]`.

abelsiqueira: This is for the .so, .f and .d files, I believe, so that the workdir remains clear.

abelsiqueira:
I made a [debug commit](https://travis-ci.org/abelsiqueira/CUTEst.jl/jobs/183541450) and indeed the problem is what's mentioned above: The `cd(tmpdir)` alone and the wrong `pwd()`. @timholy, do you wanna fix these, or can I merge locally and fix? @dpo, do you agree? I would then push directly after.

dpo:
Yes, that seems right. Thanks @abelsiqueira!

abelsiqueira:
Pushed with fix. Thanks @timholy .

timholy:
Thanks for implementing the fix! I got busy with other things.

PR #105, opened 2016-12-07
Bindeps again

abelsiqueira:
The latest build fails because of sudo ([discourse](https://discourse.julialang.org/t/installing-dependencies-error-sudo-no-tty-present-and-no-askpass-program-specified/778),
https://github.com/JuliaLang/BinDeps.jl/issues/199)), thwarting my plans of one-commands-install-all.

My current strategy is building `gsl-1.16` using BinDeps, and requiring the user to install `gfortran` separately.

The next steps are to make sure that `libgfortran` is visible, possibly inside `deps`.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,6 +1,31 @@
-# We don't use BinDeps because we couldn't make it work
-
 using Compat
+using BinDeps
+
+@BinDeps.setup
+
+libgsl = library_dependency("libgsl", aliases=["libgsl-0"])
+
+# package managers - currently with error on sudo
+#   https://discourse.julialang.org/t/installing-dependencies-error-sudo-no-tty-present-and-no-askpass-program-specified/778
+#provides(AptGet, Dict("libgsl0ldbl"=>libgsl, "libgsl0-dev" =>libgsl, "gsl-bin"=>libgsl))
+#provides(Yum, "gsl-devel", libgsl)
+#provides(Pacman, "gsl", libgsl)
+
+if is_apple()
+    if Pkg.installed("Homebrew") === nothing
+        error("Homebrew package not installed, please run Pkg.add(\"Homebrew\")")
+    end
+    using Homebrew
+    provides(Homebrew.HB, "homebrew/versions/gsl1", libgsl, os = :Darwin)
+end
+
+# build from source
+provides(Sources, URI("http://ftp.gnu.org/gnu/gsl/gsl-1.16.tar.gz"), libgsl)
+provides(BuildProcess, Autotools(libtarget = "libgsl.la"), libgsl)
+
+@BinDeps.install Dict(:libgsl => :libgsl)
+
+## BinDeps end
 
 function validate_libcutest()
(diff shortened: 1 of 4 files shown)


dpo:
If you restart the OSX jobs, you should see the same issue as in #104.

deps/build.jl
```diff
@@ -25,6 +25,34 @@ provides(BuildProcess, Autotools(libtarget = "libgsl.la"), libgsl)
...
+      end
+    end
+  else
+    found_libgfortran = true
+  end
+  if !found_libgfortran
+    error("libgfortran.so not found. Did you install it?")
+  end
+end
+
```
abelsiqueira: I was trying to have the easiest way possible to install `CUTEst.jl`. BinDeps doesn't let me install using `sudo`, so we're gonna have to ask the user to install manually with `apt-get`, `pacman` or whatever he uses. But where is `libgfortran.so` after installation is another problem, because it varies with the distribution and even the Ubuntu version, so I was trying to just search for it, instead. I think `/usr/local/lib` might sometimes not be in the library path (might be distribution specific). `usr/lib` can probably be skipped though.
However, if you think this is too much, I can remove it, and this is probably the most we can do: install gsl-1.16 for the user.

dpo: It seems to me that installing gfortran in such a way that libgfortran can be found is gfortran's responsibility. I thought BinDeps let you declare pacman, apt-get, etc., dependencies. What good are they if you can't install them? I feel this is overkill and it's reasonable to ask the user to have gfortran.

abelsiqueira: > installing gfortran in such a way that libgfortran can be found is gfortran's responsibility

I don't know who to complain about it. Possibly apt mantainers? 

> I thought BinDeps let you declare pacman, apt-get, etc., dependencies

It should, but it is broken. [See here](https://github.com/JuliaLang/BinDeps.jl/issues/199).

abelsiqueira:
The only failures are on 0.6. The `sudo ln -s` is back on Travis, and the installation is udpated.

abelsiqueira:
Rebased

dpo:
Excellent, thank you!

deps/build.jl
```diff
@@ -1,6 +1,31 @@
...
+# package managers - currently with error on sudo
+#   https://discourse.julialang.org/t/installing-dependencies-error-sudo-no-tty-present-and-no-askpass-program-specified/778
+#provides(AptGet, Dict("libgsl0ldbl"=>libgsl, "libgsl0-dev" =>libgsl, "gsl-bin"=>libgsl))
+#provides(Yum, "gsl-devel", libgsl)
+#provides(Pacman, "gsl", libgsl)
+
+if is_apple()
+    if Pkg.installed("Homebrew") === nothing
+        error("Homebrew package not installed, please run Pkg.add(\"Homebrew\")")
```
tkelman: this is unnecessary and redundant with `@osx Homebrew` in the REQUIRE file - all the packages you see this in have cargo-cult copy pasted it despite it being incorrect

PR #108, opened 2017-01-22
develop -> master

abelsiqueira:


dpo:
There's a package somewhere in our dependency tree that requires Lazy 0.11.4. That's the source of the error message on Travis with Julia 0.6. Lazy 0.11.5 fixes the issue but I'm not sure what package requires Lazy.

abelsiqueira:
JuMP uses it. But as you said, 0.11.5 fixed it. Now the error is with JuMPArray. [They haven't fixed it yet](https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/194135513), and it might be a while before they do: [here](https://github.com/JuliaOpt/JuMP.jl/issues/939#issuecomment-273482658).

dpo:
Great! The README badges still point to `develop`. After that, I guess we can delete `develop`?!

abelsiqueira:
Yes. I'm gonna push the README.

abelsiqueira:
Pushed directly.

abelsiqueira:
Btw, I don't know it that's a problem, but you merge-rebased instead of merging-ff this one.

dpo:
Thanks! I tend to prefer merge/rebase to preserve the commit history. There would only be a difference with merge-ff if we pushed anything to `master` while we were working on `develop`, but I don't think we did, did we?

abelsiqueira:
I believe not. It should be all good then.

deps/build.jl
```diff
@@ -0,0 +1,124 @@
...
+function validate_libcutest()
+  env_cutest = get(ENV, "CUTEST", "")
+  runcutest = env_cutest == "" ? "runcutest" : joinpath(env_cutest, "bin", "runcutest")
+  myarch = get(ENV, "MYARCH", "")
+  hs3 = env_cutest == "" ? "HS3.SIF" : joinpath(env_cutest, "sif", "HS3.SIF")
+  tmpdir = mktempdir()
+  try
+    cd(tmpdir) do
+      outlog = tempname()
+      errlog = tempname()
```
tkelman: should clean these up

abelsiqueira: You mean, remove the temporary files?

tkelman: yeah, when finished

Issue #113, opened 2017-01-27
Bug on cdimsh


abelsiqueira:
Some weird bug on `cdimsh` or on `HS10`. Possibly on general CUTEst, not sure.

```julia
using CUTEst
nlp = CUTEstModel("HS10")
hess(nlp, ones(2), y=[1.0]) # works
cdimsh()
hess(nlp, ones(2), y=[1.0]) # returns a different matrix
```

`HS10` has a linear objective function, but maybe not related.

abelsiqueira:
Also had problems with `cshprod` on `HS10`, but this is one harder to recreate. I'm gonna remove `HS10` from the tests, at least for now.

dpo:
I can reproduce the problem. I'll have to look if it's reproducible with Fortran.

dpo:
I can reproduce this in Fortran. Now let's try to fix it.

dpo:
Could you confirm that the latest CUTEst HEAD fixes it for you too?

abelsiqueira:
I've been using https://github.com/optimizers/cutest-mirror/ to download, can you make a new release?

abelsiqueira:
The bug was fixed, i.e., hs10 passes the consistency checks again.

I'm getting a different bug now, however.
```
sifdecoder DIXMAANJ param M=5
```
gives error on line 199. It looks like the variable `AWK` is not set.

Unfortunately I've been using an old CUTEst version, and didn't catch this sooner.

dpo:
That's probably easy to fix. A new release will take more time because the git/svn sync is broken. It'll be a few days before I can get to it.

dpo:
Fixed in HEAD. Note that the correct syntax is `sifdecoder -param M=5 DIXMAANJ`.

abelsiqueira:
While returning `hs10` to the tests, I found the error on `cshprod`.
```julia
using CUTEst

nlp = CUTEstModel("HS10")

r = hprod(nlp, nlp.meta.x0, [1.0;0.0], y=[-pi])
c = 0
err = 0.0
for i = 1:100000
  nnzr, Is, s = cshprod(nlp.meta.nvar, nlp.meta.ncon, false, nlp.meta.x0,
                        [-pi], 1, Cint[1], [1.0])
  if norm(s - r) > 1e-6
    c += 1
    err += norm(s - r)
  end
end

println("$c errors, avg $(err/c)")

finalize(nlp)
```

[...]

PR #115, opened 2017-01-27
Add many tests for specialized functions

abelsiqueira:
Due to #113, I removed `HS10` from the tests.

### DIFF
--- a/test/test_specialized.jl
+++ b/test/test_specialized.jl
@@ -299,8 +299,8 @@ function test_specinterface(nlp::CUTEstModel, comp_nlp::AbstractNLPModel)
       Jx = jac(comp_nlp, x0)
       @fact nnzj --> sum(Jx .!= 0) + nlp.meta.nvar
 
-      Wx = hess(comp_nlp, x0, y=y0)
       nnzh = cdimsh()
+      Wx = hess(comp_nlp, x0, y=y0)
       @fact nnzh --> sum(Wx .!= 0)
 
       _, _, eq, lin = cstats()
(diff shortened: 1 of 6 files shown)


test/test_specialized.jl
```diff
@@ -295,6 +295,58 @@ function test_specinterface(nlp::CUTEstModel, comp_nlp::AbstractNLPModel)
       result = zeros(J(x0)*v)
       cjprod!(nlp.meta.nvar, nlp.meta.ncon, false, false, x0, ones(nlp.meta.nvar), nlp.meta.nvar, result, nlp.meta.ncon)
 
+      nnzj = cdimsj()
+      Jx = jac(comp_nlp, x0)
+      @fact nnzj --> sum(Jx .!= 0) + nlp.meta.nvar
+
+      Wx = hess(comp_nlp, x0, y=y0)
+      nnzh = cdimsh()
```
dpo: I would have to look at the code in detail, but would guess you're supposed to call `cdimsh()` before `hess()`.

abelsiqueira: It's possible, but `CUTEstModel` constructor calls `cdimsh`.

abelsiqueira: By the way, that `hess` is for the `ADNLPModel`, for comparison with the `CUTEstModel`.

dpo: Ah, good point. So there's an issue if you call `cdimsh()` several times. That's probably reproducible in Fortran.

dpo:
Woohoo!

PR #126, opened 2017-06-10
Change https to http on pkg.julialang.org

abelsiqueira:
So that badges appear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ comparing optimization algorithms, derived from the abstract model on
 ### Stable release [![Github release](https://img.shields.io/github/release/JuliaSmoothOptimizers/CUTEst.jl.svg)](https://github.com/JuliaSmoothOptimizers/CUTEst.jl/releases/latest)
 
 - Documentation: [![Documentation](https://img.shields.io/badge/docs-stable-blue.svg)](https://JuliaSmoothOptimizers.github.io/CUTEst.jl/stable)
-- Package Evaluator: [![CUTEst.jl](https://pkg.julialang.org/badges/CUTEst_0.5.svg)](https://pkg.julialang.org/?pkg=CUTEst)
-[![CUTEst.jl](https://pkg.julialang.org/badges/CUTEst_0.6.svg)](https://pkg.julialang.org/?pkg=CUTEst)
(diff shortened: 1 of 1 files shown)


dpo:
Failed test: https://travis-ci.org/JuliaSmoothOptimizers/CUTEst.jl/jobs/241514052#L1701

abelsiqueira:
After a lot of experiments, I found out that
- ushprod sometimes returns a wrong value for BROWNDEN;
- It's partially random;
- It isn't clear what causes it. Maybe memory allocation? Something related to thread? The following code is the mwe I could come up. `@testset` is required to make it fail, strangely.
```julia
using CUTEst, Base.Test, NLPModels

nlp = CUTEstModel("BROWNDEN")
x0 = nlp.meta.x0

N = 10_000

nnzr, Ir, r = ushprod(nlp.meta.nvar, false, x0, 1, Cint[1], [1.0])
println("r = $r")
for i = 1:N
  nnzr2, Ir2, r2 = ushprod(nlp.meta.nvar, false, x0, 1, Cint[1], [1.0])
  if norm(r - r2) > 1e-6
    println("r2 = $r2, i = $i")
  end
end

@testset "Julia interface" begin
  for i = 1:N
    Hx = hess(nlp, x0)
  end
end

for i = (N+1):2N
  nnzr2, Ir2, r2 = ushprod(nlp.meta.nvar, false, x0, 1, Cint[1], [1.0])
  if norm(r - r2) > 1e-6
    println("r2 = $r2, i = $i")
  end
end

finalize(nlp)
```

dpo:
Do you mean that without `@testset`, the code behaves correctly?

abelsiqueira:
At the time, yes, but I've reduced a little more. Change all `@testset` section with
```julia
for i = 1:N
  nnzh, hval, hrow, hcol = ush(nlp.meta.nvar, x0, nlp.meta.nnzh)
  Hx = sparse(hrow, hcol, hval, nlp.meta.nvar, nlp.meta.nvar)
end
```
If you comment out `Hx = sparse...`, it won't err.

dpo:
I lost track of where things stand with this.

abelsiqueira:
The original pull request is working now, but I'm not sure it's needed anymore. I think they updated the certificates or something like that.

The random error found is now on #135.

dpo:
Is this still relevant?

abelsiqueira:
Not anymore. The PR is not useful anymore, and the error was fixed in other PRs.

PR #130, opened 2017-06-27
Change build to update CUTEst on linux

### LINKED ISSUE #129: Make `Pkg.build` reinstall cutest/sifdecode/mastsif
Better would be check if a reinstall is needed, but this would be OS specific.

Related to #128.

abelsiqueira:
Closes #129

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -59,7 +59,7 @@ if validate_libcutest()
   ispath(cutestenv) && rm(cutestenv)
 else
   install = true
-  if isfile(cutestenv)
+  if !is_linux() && isfile(cutestenv)
     include(cutestenv)
     if validate_libcutest()
       println("Valid CUTEst installation detected, no files installed.\nTo force installation, delete $cutestenv and run Pkg.build(\"CUTEst\") again.")
@@ -94,10 +94,9 @@ else
 
     @static if is_linux()
       cd(here) do
-        isdir("files") && rm("files", recursive=true)
-        mkdir("files")
+        isdir("files") || mkdir("files")
         cd("files") do
-          lnxurl = "https://raw.githubusercontent.com/abelsiqueira/linux-cutest/v0.2.1/install.sh"
+          lnxurl = "https://raw.githubusercontent.com/abelsiqueira/linux-cutest/v0.3/install.sh"
           run(`wget $lnxurl -O install.sh`)
           ENV["C_INCLUDE_PATH"] = joinpath(here, "usr", "include")
           run(`bash install.sh`)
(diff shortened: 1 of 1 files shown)


abelsiqueira:
@dpo I'm not sure if the updating is working for OSX. Otherwise, should be good to go.

deps/build.jl
```diff
@@ -54,21 +54,26 @@ end
 here = dirname(@__FILE__)
 cutestenv = joinpath(here, "cutestenv.jl")
 
+# Check if there is an external CUTEst installation
 if validate_libcutest()
+  cutest_path = get(ENV, "CUTEST", "")
+  info("External CUTEst installation found on $cutest_path")
```
dpo: Shouldn't there be a check that `cutest_path` is nonempty?

deps/build.jl
```diff
@@ -91,29 +96,32 @@ else
...
     end
+  elseif is_linux()
+    cd(here) do
+      isdir("files") || mkdir("files")
+      cd("files") do
+        if isfile("install.sh")
+          info("Updating CUTEst")
+        else
+          info("Installing CUTEst")
+        end
```
dpo: Is this `if` necessary? The action is the same in both cases.

deps/build.jl
abelsiqueira: I was thinking that `validate_libcutest` would only work if `CUTEST` existed from another installation, but I see that is not exactly the case.

abelsiqueira:
After #132, the problems that were randomly happening shouldn't happen anymore.

dpo:
What happens anyways when you say `Pkg.update()`? Is a `Pkg.build()` triggered?

deps/build.jl
```diff
@@ -54,21 +54,25 @@ end
...
-    if validate_libcutest()
-      println("Valid CUTEst installation detected, no files installed.\nTo force installation, delete $cutestenv and run Pkg.build(\"CUTEst\") again.")
-      install = false
+  @static if is_apple()
+    install = true
+    if isfile(cutestenv)
+      include(cutestenv)
+      if validate_libcutest()
+        info("Updating CUTEst")
+        Homebrew.brew(`update cutest`)
```
dpo: Should probably also update mastsif here.

deps/build.jl
```diff
@@ -91,29 +95,29 @@ else
...
-    @static if is_linux()
-      cd(here) do
-        isdir("files") && rm("files", recursive=true)
-        mkdir("files")
-        cd("files") do
-          lnxurl = "https://raw.githubusercontent.com/abelsiqueira/linux-cutest/v0.2.1/install.sh"
-          run(`wget $lnxurl -O install.sh`)
-          ENV["C_INCLUDE_PATH"] = joinpath(here, "usr", "include")
-          run(`bash install.sh`)
+        lnxurl = "https://raw.githubusercontent.com/abelsiqueira/linux-cutest/v0.3/install.sh"
```
dpo: I tagged v0.4 last night.

abelsiqueira:
From what I saw [here](https://docs.julialang.org/en/stable/stdlib/pkg/#Base.Pkg.build), `update` calls `resolve`, that calls `build`. But I haven't tested.

abelsiqueira:
Done

dpo:
Thank you!

Issue #42, opened 2015-08-07
Pros and cons of using other interfaces inside `julia_interface.jl`


abelsiqueira:
We need to evaluate the pros and cons of using the core or specialized interfaces inside `julia_interface.jl`.

From #40:

> The call chain would be
> - `objcons(nlp, x)` in the Julia interface
> - `ufn(nlp, x)` in the specialized interface
> - `ufn(io_err, n, x, f, libname)` in the core interface
> - `cutest_ufn()` in libcutest.
> 
> Right now, objcons calls cutest_ufn directly. But the intermediate functions are just passing scalars
> and pointers around, there's no copy of arrays.

dpo:
Perhaps we could just run some benchmarks on a few large problems to determine what the penalty would be.

dpo:
I'm wondering if the core interface should be exported at all. Most users will use the julia interface (I think) and if you don't want to work with NLPModels, the specialized interface is as friendly as it could be. What do you think?

As a side comment, the non`!` methods in the specialized interface should call the `!` methods. That would shorten the code and ensure better coverage.

abelsiqueira:
I think not exporting it is best.

Also, the specialized interface has a reasonable amount of functions defined.
- Two using nlp (one inplace and one not)
- Two not using nlp
- Sometimes duplicates using `Int32` for arrays.

Maybe we should remove some? I feel using `nlp` should be enough for most, and I'm not sure about `Int32` arrays.

This might take some time.

dpo:
Ok, I agree with not exporting the core interface.

Do you think a user is likely to use the specialized interface with nlp? If they use an nlp, won't they just use the julia interface?

This is by no means urgent.

abelsiqueira:
There might be (hopefully insignficant) speed differences between julia and specialized, but I think the main reason to choose one over the other are the specific functions CUTEst provides, like `ccifsg`.
It that's the case, people might still use `nlp`, but use the specialized interface.
Furthermore, I don't think they will want to manually call `sifdecoder` and `csetup`, so they'll probably create a `nlp` anyway.

dpo:
We should benchmark the different interfaces but I agree that the difference is probably not significant. I'm happy to expose the Julia and specialized interfaces.

abelsiqueira:
[...]

PR #132, opened 2017-07-30
Phase out specialized interface

abelsiqueira:
Related to #42


dpo:
Excellent, thank you! 🙏 The code base should be quite a bit easier to manage now!

PR #133, opened 2017-07-31
Make some changes for type stability.

abelsiqueira:
I've made some changes to increase type stability, but it didn't improve the timing of the functions.
Not sure if I did something wrong here. The timings are [here](https://gist.github.com/abelsiqueira/613e4794339ab537266368bdc271b2b4).

Changes previous `cons` to `consjac` that returns the constraints and Jacobian, while the currenty `cons` only return the constraints, like `NLPModels`.

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -64,7 +64,6 @@ CUTEstException(info :: Integer) = CUTEstException(convert(Int32, info));
 
 macro cutest_error()  # Handle nonzero exit codes.
   esc(:(io_err[1] > 0 && throw(CUTEstException(io_err[1]))))
-  nothing
 end
 
 include("core_interface.jl")
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -64,6 +64,7 @@ CUTEstException(info :: Integer) = CUTEstException(convert(Int32, info));
 
 macro cutest_error()  # Handle nonzero exit codes.
   esc(:(io_err[1] > 0 && throw(CUTEstException(io_err[1]))))
+  nothing
 end
 
 include("core_interface.jl")
--- a/src/julia_interface.jl
+++ b/src/julia_interface.jl
@@ -1,5 +1,5 @@
-export objcons, objgrad, obj, grad, grad!,
-       cons_coord, cons, cons!,
+export objcons, objgrad, objgrad!, obj, grad, grad!,
+       cons_coord, consjac, cons, cons!,
        jac_coord, jac, jprod, jprod!, jtprod, jtprod!,
        hess_coord, hess, hess_op, hprod, hprod!
 
@@ -76,7 +76,7 @@ function objgrad(nlp :: CUTEstModel, x :: Array{Float64,1}, grad :: Bool = true)
   grad && (nlp.counters.neval_grad += 1)
   @cutest_error
 
-  return grad ? (f[1], g) : f[1];
+  return f[1], g
 end
 
 @doc (@doc NLPModels.objgrad!)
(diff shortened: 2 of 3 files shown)


src/CUTEst.jl
```diff
@@ -64,6 +64,7 @@ CUTEstException(info :: Integer) = CUTEstException(convert(Int32, info));
 
 macro cutest_error()  # Handle nonzero exit codes.
   esc(:(io_err[1] > 0 && throw(CUTEstException(io_err[1]))))
+  nothing
```
dpo: I don't think this is needed. It will be substituted everywhere we use the macro and will do nothing.

src/julia_interface.jl
```diff
@@ -76,7 +76,7 @@ function objgrad(nlp :: CUTEstModel, x :: Array{Float64,1}, grad :: Bool = true)
   grad && (nlp.counters.neval_grad += 1)
   @cutest_error
 
-  return grad ? (f[1], g) : f[1];
+  return f[1], g
```
dpo: Should we consider that `grad` is always `true` when the user calls `objgrad`? Otherwise, they would just call `obj` and `grad` separately.

src/julia_interface.jl
```diff
@@ -139,16 +139,15 @@ Computes the constraint vector and, if `jac` is `true`, the Jacobian in
 coordinate format.
 Usage:
 
-    c, jrow, jcol, jval = cons_coord(nlp, x, true)
-    c = cons_coord(nlp, x, false)
+    c, jrow, jcol, jval = cons_coord(nlp, x, jac)
```
dpo: Similarly here, if the user calls `cons_coord()` instead of `cons()`, then we should consider that `jac = true`.

src/julia_interface.jl
```diff
@@ -168,36 +167,31 @@ function cons_coord(nlp :: CUTEstModel, x :: Array{Float64,1}, jac :: Bool)
...
-    return cons_coord(nlp, x, jac);
-  end
+function consjac(nlp :: CUTEstModel, x :: Array{Float64,1}, jac :: Bool)
+  c, jrow, jcol, jval = cons_coord(nlp, x, jac);
+  return c, sparse(jrow, jcol, jval, nlp.meta.ncon, nlp.meta.nvar)
 end
 
 @doc (@doc NLPModels.cons)
-cons(nlp :: CUTEstModel, x :: Array{Float64,1}) = cons_coord(nlp, x, false);
+cons(nlp :: CUTEstModel, x :: Array{Float64,1}) = cons_coord(nlp, x, false)[1]
```
dpo: For `cons`, we should call `cutest_cfn`.

src/julia_interface.jl
dpo: That will allow us to simplify some in-place methods, e.g.,
```julia
function objgrad(nlp :: CUTEstModel, x :: Array{Float64,1})
  g = Array{Float64}(nvar);
  objgrad!(nlp, x, g)
end
```

abelsiqueira:
Updated with suggestions.

dpo:
Nice! That seems much more concise!

src/julia_interface.jl
```diff
@@ -1,31 +1,14 @@
...
-    ccall(dlsym(cutest_lib, :cutest_ufn_), Void,
-                (Ptr{Int32}, Ptr{Int32}, Ptr{Float64}, Ptr{Float64}),
-                 io_err,  &nvar,    x,         f);
-  end
-  nlp.counters.neval_obj += 1
-  @cutest_error
-
-  return f[1], c
+  c = Array{Float64}(nlp.meta.ncon);
+  objcons!(nlp, x, c)
```
dpo: 🎉

Issue #134, opened 2017-08-01
Linuxbrew


dpo:
[Linuxbrew](http://linuxbrew.sh) has been picking up steam. Recently, I forked Homebrew.jl and made [Linuxbrew.jl](https://github.com/dpo/Linuxbrew.jl). Some tests are currently failing and I'm not yet sure why but it seems inconsequential. I can install formulae, including precompiled bottles.

I just installed and tested CUTEst with Linuxbrew and all works well. I also tested the installation within Julia using this modified `build.jl`: https://gist.github.com/3de7180903daf6af3004cc50b119864d

Lo and behold: it works!

`Pkg.test("CUTEst")` is a drag because dependencies like IPOPT build lots of redundant stuff, but hopefully, that can be improved.

It would be great if you could give it a try.

dpo:
```julia
Test Summary:   | Pass  Total
Julia interface |   14     14
Core interface stress test... passed
Test Summary:  | Pass  Total
Core interface |   11     11
Finalizing
Testing interfaces on problem hs5
Julia interface stress test... Passed
Test Summary:   | Pass  Total
Julia interface |   14     14
Core interface stress test... passed
Test Summary:  | Pass  Total
Core interface |   11     11
Finalizing
Testing interfaces on problem hs6
Julia interface stress test... Passed
Test Summary:   | Pass  Total
Julia interface |   35     35
Core interface stress test... passed
Test Summary:  | Pass  Total
Core interface |   63     63
Finalizing
Testing interfaces on problem hs11
Julia interface stress test... Passed
Test Summary:   | Pass  Total
Julia interface |   35     35
Core interface stress test... passed
Test Summary:  | Pass  Total
Core interface |   63     63
Finalizing
Testing interfaces on problem hs14
Julia interface stress test... Passed
Test Summary:   | Pass  Total
Julia interface |   35     35
Core interface stress test... passed
Test Summary:  | Pass  Total
Core interface |   68     68
# ... etc.
Test Summary:         | Pass  Total
CUTEst selection tool | 1395   1395
```

abelsiqueira:
First thing I tried was installing over what I already have. It downloaded `homebrew-core`, `homebrew-julia`, then failed on `brew update cutest` with error:
```
Error: This command updates brew itself, and does not take formula names.
Use 'brew upgrade <formula>'.
=======================================[ ERROR: CUTEst ]=======================================

[...]

PR #136, opened 2017-08-01
Linuxbrew

dpo:
cf #134

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,16 +9,15 @@ matrix:
       julia: 0.6
       osx_image: xcode7.3
 
-addons:
-  apt_packages:
-    - gfortran
-
 branches:
   only:
     - master
 
 before_install:
-  - if [ `uname` == "Linux" ]; then unset DY_LIBRARY_PATH; fi
+  - if [ `uname` == "Linux" ]; then
+      unset DY_LIBRARY_PATH;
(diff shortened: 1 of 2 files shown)


dpo:
@abelsiqueira This build succeeded but I had to pre-install gfortran on the Linux VM. I'll see if it's possible to use Linuxbrew's own gfortran.

abelsiqueira:
Maybe `Linuxbrew.add("patchelf")` could be a part of the `build.jl`.

abelsiqueira:
What do we do here? I don't use Linuxbrew for anything - it's not so useful on linux - but it uses the same bottle as homebrew, which should be easier to maintain. The hardest thing was finding `libgfortran.so`, because it's different on different Ubuntu versions, and on Archlinux, but I think the latest `linux-cutest` and the updates on the installation have mostly fixed that.

dpo:
The one advantage I see to Linuxbrew is that it's useful if you don't have admin privileges (and you can't `apt-get install` something). But it's not perfect and sometimes it's a pain. Why do you say it's not so useful?

abelsiqueira:
Most distributions come with a package manager (`apt-get`, or in my case `pacman`), which generally has most of what the users need. Possibly the niche packages, but I haven't used it. But you're right about the privileges.

PR #138, opened 2017-11-01
hess and jac indexes are in Int64

abelsiqueira:
I was testing `cholfact` with `hess`, and it didn't work because `hess` currently returns a `SparseMatrixCSC{Float64,Int32}`, and the `Int32` needs to be `Int64`. Changing only `hess` and not `hess_coord` should also work, what do you think?

Failing code:
```julia
Hx = Symmetric(hess(nlp, x), :L)
F = cholfact(Hx)
```


dpo:
Looks good to me, thanks! I think the fix for Travis/OSX should look like this: https://github.com/liballeg/allegro5/pull/855/files.

abelsiqueira:
It seems that [Homebrew.jl isn't being maintained for 0.5 anymore](https://github.com/JuliaPackaging/Homebrew.jl/issues/209).

dpo:
Thanks. Let's leave it for now and we'll drop 0.5 support at some point.

PR #142, opened 2017-12-03
Update installation to keep a link to libgfortran.so on deps/

abelsiqueira:
Related to https://github.com/abelsiqueira/linux-cutest/issues/11

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -23,6 +23,7 @@ const automat = "AUTOMAT.d";
 const funit   = convert(Int32, 42);
 @static is_apple() ? (const linker = "gfortran") : (const linker = "ld")
 @static is_apple() ? (const sh_flags = ["-dynamiclib", "-undefined", "dynamic_lookup"]) : (const sh_flags = ["-shared"]);
+@static is_apple() ? (const libgfortran = []) : (const libgfortran = ["../../usr/lib/libgfortran.so"])
 
 type CUTEstException <: Exception
   info :: Int32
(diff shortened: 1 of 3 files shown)


abelsiqueira:
@dpo Does this work for your server?

dpo:
I can't try now as I'm traveling. I'll have to do it in a few days. If it works on TravisCI, then it should be ok. However, shouldn't we use an absolute path instead of `../../` in the location of `libgfortran.so`? Something like `joinpath(dirname(@__FILE__), "..", "usr", "lib", "libgfortran.so")`?!

src/CUTEst.jl
```diff
@@ -23,6 +23,7 @@ const automat = "AUTOMAT.d";
 const funit   = convert(Int32, 42);
 @static is_apple() ? (const linker = "gfortran") : (const linker = "ld")
 @static is_apple() ? (const sh_flags = ["-dynamiclib", "-undefined", "dynamic_lookup"]) : (const sh_flags = ["-shared"]);
+@static is_apple() ? (const libgfortran = []) : (const libgfortran = [joinpath(@__DIR__, "..", "deps", "usr", "lib", "libgfortran.so")])
```
dpo: I was testing this and got distracted by #144. Is the `.so` intentional here, or should it be `$(Libdl.dlext)`?

abelsiqueira: Intentional, because it's specific for Linux, and also because a .so is created by linux-cutest.

dpo:
Thanks.

PR #143, opened 2017-12-21
Documentation fix

abelsiqueira:
Documentation was not being generated due to changes on Documenter.jl and version change. Should work after this.


dpo:
Thank you!

PR #144, opened 2017-12-29
catch brew upgrade error

dpo:
`brew upgrade` returns an error if the latest version of a formula is already installed. `Pkg.build("CUTEst")` fails if we don't catch that error.


abelsiqueira:
Thanks.

PR #145, opened 2018-01-02
Make OUTSDIF.d problem dependent

anriseth:
This allows different `CUTEstModels` to be run in parallel.

I don't know that `AUTOMAT.d` does, but I'm storing one for each model in case someone knows what to do with them.

Fixes https://github.com/JuliaSmoothOptimizers/CUTEst.jl/issues/51#issuecomment-354757852

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,20 +45,48 @@ After installing, you can create instances of
 [NLPModels](https://github.com/JuliaSmoothOptimizers/NLPModels.jl) models, with
 the name `CUTEstModel`:
 
-````JULIA
+```jl
 using CUTEst
 
 nlp = CUTEstModel("BYRDSPHR");
 print(nlp);
-````
+```
 
 This model accepts the same functions as the other NLPModels, for instance
 
-````
+```jl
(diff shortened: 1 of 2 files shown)


abelsiqueira:
I tried running the code on #51 using `julia -p 2`, but it didn't work for me. Am I doing something wrong?

anriseth:
You need to decode each of the problems in serial first:
```jl
addprocs(2)
@everywhere using CUTEst
@everywhere function opt(name, decode)
    nlp = CUTEstModel(name;decode=decode)
    retval = obj(nlp,nlp.meta.x0)
    #retval = nlp.meta.nvar
    finalize(nlp)
    retval
end

probs = sort(CUTEst.select(contype="unc",max_var=5))
map(x->opt(x,true), probs)
pmap(x->opt(x,false), probs)
```

abelsiqueira:
Yes, it worked now, thanks.

anriseth:
I've included an example for the README. Let me know if you want it elsewhere.

abelsiqueira:
Thank you

Issue #120, opened 2017-03-07
Create select that returns a generator expression


abelsiqueira:
Equivalent to
`(CUTEstModel(p) for p in CUTEst.select(...))`

PR #146, opened 2018-01-02
Return Array{String} from select

anriseth:
This may just be my own preference, but I expected the array returned from CUTEst.select to be of type Array{String,1}


abelsiqueira:
That's correct. Thanks.

abelsiqueira:
We hope to have a generator instead, though (#120)

PR #152, opened 2018-03-06
Update Zenodo to v0.3.2

abelsiqueira:
@dpo Do you have experience with Zenodo? Are we supposed to update the version every time before a new release?


dpo:
I have no idea.

pkofod:
You just tag a new version and it updates shortly after

abelsiqueira:
Thanks @pkofod, I think I've got it working now.

PR #155, opened 2018-04-11
fix location of libgfortran on linux

dpo:
See https://github.com/abelsiqueira/linux-cutest/issues/14#issuecomment-380606665.


dpo:
@abelsiqueira Why does the linux installer need to create a symbolic link to `libgfortran` if its location is defined in `CUTEst.jl`?

abelsiqueira:
Workaround not knowing about `gfortran --print-file` and to have `libgfortran.so` on a known place. With this PR I can probably leave `libgfortran.so` whenever and not link.

abelsiqueira:
Thanks

PR #156, opened 2018-04-12
Use linux-cutest-0.4.0

abelsiqueira:
New linux-cutest-0.4.0 doesn't create a link to libgfortran, since
it is no longer necessary


dpo:
Thank you.

PR #158, opened 2018-07-18
fix install of GSL

dpo:
- fully qualify CUTEst tap
- only install GSL if another installation is not found


dpo:
The issue is https://github.com/JuliaPackaging/Homebrew.jl/issues/231. It works for me locally.

abelsiqueira:
Thanks

PR #160, opened 2018-09-07
Last release with MathProgBase and NLPModels 0.4.3

abelsiqueira:


dpo:
CUTEst build error on macOS due to https://github.com/Homebrew/homebrew-core/issues/27751.

abelsiqueira:
It looks like there's nothing we can do for now. Since the change is only on upper bounds, I'll merge this.

PR #168, opened 2018-11-14
fail fast if file does not exist

### LINKED ISSUE #165: Fail more elegantly
When opening a problem that doesn't exist, we currently get
```julia
julia> model = CUTEstModel("BOYD1.SIF")
gfortran: error: ELFUN.f: No such file or directory
gfortran: error: EXTER.f: No such file or directory
gfortran: error: GROUP.f: No such file or directory
gfortran: error: RANGE.f: No such file or directory
gfortran: fatal error: no input files
compilation terminated.
ERROR: failed process: Process(`gfortran -c -fPIC ELFUN.f EXTER.f GROUP.f RANGE.f`, ProcessExited(1)) [1]
```
We should fail more elegantly (and quickly) when the file doesn't exist.

dpo:
fixes #165

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -81,6 +81,14 @@ Example:
 function sifdecoder(name :: String, args...; verbose :: Bool=false,
                     outsdif :: String="OUTSDIF_$name.d",
                     automat :: String="AUTOMAT_$name.d")
+
+  if length(name) < 4 || name[end-3:end] != ".SIF"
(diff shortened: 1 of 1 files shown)


abelsiqueira:
I've tested the "Resolve Conflict" GitHub feature, but it creates a new commit, so I'm gonna merge by hand.

abelsiqueira:
Merged by hand. Thanks you.

PR #169, opened 2018-11-16
allow user to specify ordering preference

dpo:
There's an extra option in CUTEst, which consists in setting `e_order`, etc., to 2, and places equalities/linear constraints/linear variables *after* inequalities/nonlinear constraints/nonlinear variables. Should we support it, or is this sufficient?


abelsiqueira:
I think it's sufficient.

abelsiqueira:
Should we add a test? Seems straightforward that is should work.

dpo:
I'm not sure what to test.

PR #173, opened 2019-01-27
Allow AbstractVector in interface

abelsiqueira:
Uses `test_view_array` from `NLPModels.jl`.

### DIFF
--- a/src/julia_interface.jl
+++ b/src/julia_interface.jl
@@ -3,7 +3,7 @@ export cons_coord, consjac
 using NLPModels, SparseArrays
 
 @doc (@doc NLPModels.objcons)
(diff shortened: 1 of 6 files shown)


src/julia_interface.jl
```diff
@@ -30,8 +30,24 @@ function NLPModels.objcons!(nlp :: CUTEstModel, x :: Array{Float64,1}, c :: Arra
   return f[1], c
 end
 
+function NLPModels.objcons!(nlp :: CUTEstModel, x :: AbstractVector, c :: Array{Float64,1})
+  objcons!(nlp, Vector{Float64}(x), c)
```
dpo: Good idea!

dpo:
Thank you!

Issue #176, opened 2019-03-02
Test failures in SIF collection


dpo:
I'm seeing test failures of the form
```
CUTEst selection tool: Test Failed at /Users/dpo/dev/julia/JSO/CUTEst.jl/test/test_select.jl:20
  Expression: np == length(CUTEst.select())
   Evaluated: 1305 == 1315
Stacktrace:
 [1] macro expansion at /Users/dpo/dev/julia/JSO/CUTEst.jl/test/test_select.jl:20 [inlined]
 [2] macro expansion at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
 [3] top-level scope at /Users/dpo/dev/julia/JSO/CUTEst.jl/test/test_select.jl:18
Variables: Test Failed at /Users/dpo/dev/julia/JSO/CUTEst.jl/test/test_select.jl:29
  Expression: np == n + length(selection)
   Evaluated: 1305 == 1315
Stacktrace:
 [1] macro expansion at /Users/dpo/dev/julia/JSO/CUTEst.jl/test/test_select.jl:29 [inlined]
 [2] macro expansion at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
 [3] macro expansion at /Users/dpo/dev/julia/JSO/CUTEst.jl/test/test_select.jl:22 [inlined]
 [4] macro expansion at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
 [5] top-level scope at /Users/dpo/dev/julia/JSO/CUTEst.jl/test/test_select.jl:18
```
They are probably due to an updated SIF collection?

abelsiqueira:
Yes, the internal selection tool uses a different list of problems from
MASTSIF, so it returns an error. Maybe there is no need for such a check.

dpo:
Fixed in #209.

PR #177, opened 2019-03-02
draft single precision support

dpo:
These are preliminary changes to support calling CUTEst in single precision. I haven't yet added single-precision tests. I see test failures locally that are probably due to an updated SIF collection (see #176).

The changes look a bit messy because I took the docstrings out of the `@eval` blocks.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -29,12 +29,14 @@ function validate_libcutest()
   tmpdir = mktempdir()
   try
     cd(tmpdir) do
-      outlog = tempname()
-      errlog = tempname()
-      run(pipeline(`$runcutest -p genc -D $hs3`, stdout=outlog, stderr=errlog))
-      print(read(errlog, String))
-      rm(outlog, force=true)
-      rm(errlog, force=true)
+      for prec in ("", "--single")
+        outlog = tempname()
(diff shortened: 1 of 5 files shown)


dpo:
@abelsiqueira Does your CUTEst installer also install the single precision library? If not, we're going to need it.

abelsiqueira:
Updating here: https://github.com/abelsiqueira/linux-cutest/pull/15

dpo:
`https://raw.githubusercontent.com/abelsiqueira/linux-cutest/v0.4.2/install.sh` still says `VERSION=0.4.1`. Not sure if it matters. I'll update the version number in `build.jl` and see if it works.

dpo:
It's looking good! The Linux tests pass. Now I'll have to add single-precision tests.

abelsiqueira:
Looking very good, thanks.

abelsiqueira:
[This](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/master/test/multiple-precision.jl) is what's being tested in `NLPModels.jl`, perhaps it can be reused.

dpo:
Yes, those are definitely important tests.

amontoison:
Done in #366

PR #179, opened 2019-03-13
Single precision update

abelsiqueira:
Tried updating #177, but there are some errors. Maybe `ELFUN.f` and similar are not multipre precision?

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -33,9 +33,13 @@ function validate_libcutest()
         outlog = tempname()
         errlog = tempname()
         run(pipeline(`$runcutest -p genc $prec -D $hs3`, stdout=outlog, stderr=errlog))
-        print(read(errlog, String))
+        errstr = read(errlog, String)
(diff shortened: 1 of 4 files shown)


abelsiqueira:
Sorry, forgot to say I had segfaults too. The simplest segfault was something like
```julia
nlp = CUTEstModel("ROSENBR")
io_err = Cint[0]
nvar = 2
x = Cfloat[-1.2; 1.0]
f = Cfloat[0]
ccall(dlsym(CUTEst.cutest_lib_single, :cutest_ufn_), Nothing,
      (Ptr{Int32}, Ref{Int32}, Ptr{Cfloat}, Ptr{Cfloat}),
      io_err, nvar, x, f)
```

dpo:
Same here. Decoding the problem with `--single` has no effect. I don't think the problem is in the Julia interface.

amontoison:
@abelsiqueira 
I merged your modifications in #371

PR #182, opened 2019-03-17
Implement coord and structure functions

abelsiqueira:


src/julia_interface.jl
```diff
@@ -108,6 +108,39 @@ function NLPModels.grad!(nlp :: CUTEstModel, x :: AbstractVector, g :: AbstractV
...
+
+  - nlp:  [IN] CUTEstModel
+  - x:    [IN] Array{Float64, 1}
+  - c:    [OUT] Array{Float64, 1}
+  - jrow: [OUT] Array{Int32, 1}
+  - jcol: [OUT] Array{Int32, 1}
+  - jval: [OUT] Array{Float64, 1}
+"""
+function cons_coord!(nlp :: CUTEstModel, x :: AbstractVector, c :: AbstractVector, rows :: AbstractVector{<:Integer}, cols :: AbstractVector{<:Integer}, vals :: AbstractVector)
+  nvar = nlp.meta.nvar;
```
dpo: We don't need those semicolons. Eventually, we should remove them all (in PRs that only do that). They're not necessary and clutter up the code.

src/julia_interface.jl
```diff
@@ -108,6 +108,39 @@ function NLPModels.grad!(nlp :: CUTEstModel, x :: AbstractVector, g :: AbstractV
...
+    cons_coord!(nlp, x, c, rows, cols, vals)
+
+  - nlp:  [IN] CUTEstModel
+  - x:    [IN] Array{Float64, 1}
+  - c:    [OUT] Array{Float64, 1}
+  - jrow: [OUT] Array{Int32, 1}
+  - jcol: [OUT] Array{Int32, 1}
+  - jval: [OUT] Array{Float64, 1}
+"""
+function cons_coord!(nlp :: CUTEstModel, x :: AbstractVector, c :: AbstractVector, rows :: AbstractVector{<:Integer}, cols :: AbstractVector{<:Integer}, vals :: AbstractVector)
```
dpo: I don't think we want to allow other `Integer` types here. CUTEst only accepts `Cint`.

src/julia_interface.jl
abelsiqueira: #183

dpo: Right, but let's no commit new code with semicolons.

src/julia_interface.jl
abelsiqueira: I've changed to use `Int32`, and convert other `Integer` to `Int32`.

dpo:
Great, thank you!

abelsiqueira:
Now updated, removed the `;`, changed `Integer` functions, and updated the `docs` to pass the tests. Should be ready to go.

PR #185, opened 2019-03-27
implement in-place hess_structure!

dpo:
Needs https://github.com/JuliaSmoothOptimizers/NLPModels.jl/pull/183.

### DIFF
--- a/src/julia_interface.jl
+++ b/src/julia_interface.jl
@@ -319,13 +319,20 @@ end
 
 @doc (@doc NLPModels.hess_structure)
 function NLPModels.hess_structure(nlp :: CUTEstModel)
+  nnzh = nlp.meta.nnzh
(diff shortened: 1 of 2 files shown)


src/julia_interface.jl
```diff
@@ -338,8 +345,6 @@ function NLPModels.hess_structure(nlp :: CUTEstModel)
                 io_err, nvar, this_nnzh, nnzh, cols, rows)
     @cutest_error
   end
-
-  return rows, cols
```
abelsiqueira: We should probably still return `rows` and `cols`.

dpo: Done, thanks.

dpo:
Completely forgot I had opened this. Your request is legitimate. In fact, lots of methods require similar changes because we must ensure that we pass contiguous arrays to CUTEst. Currently, we don't check that.

PR #187, opened 2019-07-06
in-place structure methods

dpo:
requires https://github.com/JuliaSmoothOptimizers/NLPModels.jl/pull/188


dpo:
I've no idea why Documenter is complaining. I've temporarily disabled `strict=true` in `make.jl`.

@abelsiqueira Any ideas?

PR #189, opened 2019-07-07
Remove NLPModels docstrings and add link to API in the docs

abelsiqueira:
The docstrings were duplicated.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -4,7 +4,7 @@ makedocs(
   modules = [CUTEst],
   doctest = true,
   linkcheck = true,
-  # strict = true,
+  strict = true,
   assets = ["assets/style.css"],
   format = Documenter.HTML(
               prettyurls = get(ENV, "CI", nothing) == "true"
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -2,35 +2,11 @@
 
 ## NLPModels API
 
-```@docs
-obj
-grad
-grad!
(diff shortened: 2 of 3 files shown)


dpo:
Yes, that was a bit weird. Thank you. Confusingly, Documenter complained about *missing* docstrings!

abelsiqueira:
It was because `hess_structure!` didn't have a docstring, but it was included in `api.md`. The solution would be adding `@doc (@doc NLPModels.hess_structure!)` in the function definition. I then realized that removing the docstrings altogether would be better.

dpo:
It [did](https://github.com/JuliaSmoothOptimizers/CUTEst.jl/commit/c7135e403699c769afaf4fe9bf71adeaee772378#diff-582c252a3e3e6631014cd1bc8d0ccf3eR349) (line 349). Other methods were similar, but Documenter wasn't complaining about them.

abelsiqueira:
I see, I was looking at [this line](https://github.com/JuliaSmoothOptimizers/CUTEst.jl/commit/c7135e403699c769afaf4fe9bf71adeaee772378#diff-582c252a3e3e6631014cd1bc8d0ccf3eR327) (327).

It might be that the problem was that it didn't have `NLPModels.` preceding the function name. This also begs the question if the function is exported.

dpo:
I wish Documenter (and Julia in general) had good error messages...

PR #191, opened 2019-07-07
Add NLPModels. to hess_structure!

abelsiqueira:
Without `NLPModels.` the function being called is the one right below in line 332. This can be verified with `@which`.


dpo:
Thanks. I was fixing that in #188 but I can leave it out if you prefer.

abelsiqueira:
Better to add in #188.

PR #196, opened 2019-07-15
allow command that does not decode a problem

dpo:
The command `sifdecoder("LUBRIFC", "-show")` currently fails because it does not decode a problem, but we still try to compile `ELFUN.f`, etc. This change checks that a Fortran file is produced before trying to compile.

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -72,6 +72,19 @@ include("core_interface.jl")
 include("julia_interface.jl")
 include("classification.jl")
 
+function delete_temp_files()
+  for f in ("ELFUN", "EXTER", "GROUP", "RANGE")
+    for ext in ("f", "o")
+      fname = "$f.$ext"
+      isfile(fname) && rm(fname, force=true)
+    end
+  end
+  for f in ("OUTSDIF.d", "AUTOMAT.d")
+    isfile(f) && rm(f, force=true)
+  end
+  nothing
+end
+
 """Decode problem and build shared library.
 
 Optional arguments are passed directly to the SIF decoder.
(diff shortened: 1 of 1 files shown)


dpo:
With this PR:
```
julia> sifdecoder("LUBRIFC", "-show", verbose=true)
NN=10	(IE)			comment:  n = 151 original value
NN=50	(IE)			comment:  n = 751
NN=250	(IE)	-default value-	comment:  n = 3751
```

src/CUTEst.jl
```diff
@@ -98,17 +98,20 @@ function sifdecoder(name :: String, args...; verbose :: Bool=false,
   outlog = tempname()
   errlog = tempname()
   cd(ENV["cutest-problems"]) do
+    run(`rm -f ELFUN.[fo] EXTER.[fo] GROUP.[fo] RANGE.[fo] OUTSDIF.d AUTOMAT.d`)
```
abelsiqueira: This construction doesn't work on Linux, for security reasons, apparently, but luckily, looking [here](https://docs.julialang.org/en/v1/manual/running-external-programs/index.html), I've found that we can make
```
fnames = ["ELFUN", "EXTER", "GROUP", "RANGE"]
exts = ["f", "o"]
run(`rm -f $fnames.$exts OUTSDIF.d AUTOMAT.d`)
```
i.e., interpolating with an array simulates the `{a,b,c}` shell command.

src/CUTEst.jl
```diff
@@ -98,17 +98,20 @@ function sifdecoder(name :: String, args...; verbose :: Bool=false,
...
     verbose && println(read(outlog, String))
 
-    run(`gfortran -c -fPIC ELFUN.f EXTER.f GROUP.f RANGE.f`)
-    run(`$linker $sh_flags -o $libname.$(Libdl.dlext) ELFUN.o EXTER.o GROUP.o RANGE.o $libpath $libgfortran`)
-    run(`mv OUTSDIF.d $outsdif`)
-    run(`mv AUTOMAT.d $automat`)
-    run(`rm ELFUN.f EXTER.f GROUP.f RANGE.f ELFUN.o EXTER.o GROUP.o RANGE.o`)
-    global cutest_lib = Libdl.dlopen(libname,
-      Libdl.RTLD_NOW | Libdl.RTLD_DEEPBIND | Libdl.RTLD_GLOBAL)
+    if isfile("ELFUN.F")
```
abelsiqueira: `F` -> `f`.

dpo: Done.

src/CUTEst.jl
dpo: Thanks. I decided to add a separate function to delete temporary files. It works on macOS. Let's see if Linux is satisfied.

abelsiqueira:
Thank you

dpo:
@ppascal97 This should help with some of the problems you were having.

PR #198, opened 2019-07-15
store sparsity structure inside CUTEstModel

dpo:
jac_coord! and hess_coord! should not assume that the rows and cols
index vectors are well defined and should not try to access them


abelsiqueira:
Thank you. I think we can release now, right?

dpo:
@PEPERSO Could you test this with NCL?

dpo:
@abelsiqueira It should be good, but let's test it to make sure.

PEPERSO:
@dpo All my test with ```NCL``` passed, and I made sure to use ```hess_structure!``` and ```jac_structure!``` in my model, so it should be alright.

dpo:
Excellent, thank you!

PR #201, opened 2019-08-10
Add BinDeps to deps

timholy:
Can't `Pkg.build("CUTEst")` without it


abelsiqueira:
Thank you. Tests failed because sometimes some downloads don't work and I have to manually restart the jobs.

PR #204, opened 2019-11-18
Upgrade nlpmodels

dpo:


src/julia_interface.jl
```diff
@@ -344,7 +344,7 @@ function NLPModels.hess_structure!(nlp :: CUTEstModel, rows :: AbstractVector{<:
   return rows, cols
 end
 
-function NLPModels.hess_coord!(nlp :: CUTEstModel, x :: Vector{Float64}, rows :: Vector{Int32}, cols :: Vector{Int32}, vals :: Vector{Float64}; y :: Vector{Float64}=zeros(nlp.meta.ncon), obj_weight :: Float64=1.0)
+function NLPModels.hess_coord!(nlp :: CUTEstModel, x :: Vector{Float64}, y :: Vector{Float64}, vals :: Vector{Float64}; obj_weight :: Float64=1.0)
```
abelsiqueira: We can define `hess_coord!(nlp, x, vals)` as well. The default creates `y=zeros(nvar)`

src/julia_interface.jl
```diff
@@ -376,17 +376,21 @@ function NLPModels.hess_coord!(nlp :: CUTEstModel, x :: Vector{Float64}, rows ::
...
-  return rows, cols, vals
+  return vals
+end
+
+function NLPModels.hess_coord!(nlp :: CUTEstModel, x :: AbstractVector, vals :: AbstractVector; obj_weight :: Float64=1.0)
+  hess_coord!(nlp, x, zeros(nlp.meta.ncon), vals; obj_weight=obj_weight)
 end
 
-function NLPModels.hprod!(nlp :: CUTEstModel, x :: Vector{Float64}, v :: Vector{Float64}, hv :: Vector{Float64}; y :: Vector{Float64}=zeros(nlp.meta.ncon), obj_weight :: Float64=1.0)
+function NLPModels.hprod!(nlp :: CUTEstModel, x :: Vector{Float64}, y :: Vector{Float64}, v :: Vector{Float64}, hv :: Vector{Float64}; obj_weight :: Float64=1.0)
```
abelsiqueira: We can also define `hprod!(nlp, x, v)`.

dpo: I defined them further down.

dpo:
I'll improve coverage tonight.

src/julia_interface.jl
abelsiqueira: We can define a more specialized one that doesn't allocate `zeros`.

dpo: See lin 436. There's no other way with CUTEst I think. There's no special function to only compute the product with the Hessian of the objective. You can't use ush here.

abelsiqueira: Wouldn't the case without `y` would be equivalent to the case with `ncon = 0`? You can use `cutest_uhprod_`?

dpo: When I tried on a constrained problem, it gave incorrect results. I don't think it's safe to call ufn when m > 0. That's why I create a vector of zero multipliers. To my knowledge, there's no other way.

abelsiqueira: I see, thanks for explaining.

dpo:
Not sure I trust what Coveralls says. The tests are definitely calling `jac_structure` and `hess_structure`.

PR #209, opened 2020-01-21
Use precompiled binaries for julia ≥ 1.3

dpo:


abelsiqueira:
:fireworks:

Project.toml
```diff
@@ -3,17 +3,17 @@ uuid = "1b53aba6-35b6-5f92-a507-53c67d53f819"
...
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NLPModels = "a4795742-8479-5a88-8948-cc11e1c8c1a6"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
-Combinatorics = "1.0"
 CUTEst_jll = "^0.1.0"
+Combinatorics = "1.0"
```
dpo: ?

abelsiqueira: I just ran `pkg> up` and committed everything that changed. Apparently, it likes this order.

dpo:
Rebased so the commit history is a bit cleaner.

.travis.yml
```diff
@@ -26,17 +29,20 @@ branches:
```
abelsiqueira: I believe we can remove `apt_packages: gfortran` as well, since `CUTEst_jll` uses the one shipped with Julia.

Project.toml
```diff
@@ -3,24 +3,22 @@ uuid = "1b53aba6-35b6-5f92-a507-53c67d53f819"
...
+CUTEst_jll = "^0.1.0"
 Combinatorics = "1.0"
 DataStructures = "0.17"
-Homebrew = "0.4, 0.5, 0.6, 0.7"
 JSON = "0.8, 0.9, 0.10, 0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17, 0.18, 0.19, 0.20, 0.21"
-Krylov = "0.4"
-NLPModels = "0.10, 1"
-julia = "^1.0.0"
+Krylov = "0.2.0, 0.3.0, 0.4.0"
+NLPModels = "0.10.0, 1"
```
abelsiqueira: I think we should remove ",1", because there may be an `NLPModels` 0.11 before 1.0.

.travis.yml
dpo: We still need gfortran to compile decoded problems.

dpo: The jll only links the executables and librarie to libgfortran. We don't ship gfortran.

Project.toml
dpo: I suggest we add it in when we release it (we'll have to do it everywhere).

dpo:
🎆

PR #211, opened 2020-01-26
fix path to problems in consistency test

dpo:


dpo:
keep track of the breakage reported in JuliaSmoothOptimizers/NLPModels.jl#236.

abelsiqueira:
Now fixed.

dpo:
Thanks.

PR #220, opened 2020-03-29
Add Julia 1.4 to CI

abelsiqueira:


.travis.yml
```diff
@@ -41,6 +42,6 @@ jobs:
       after_success: skip
 
 after_success:
-  - julia -e 'if Sys.islinux() && string(VERSION)[1:3] == "1.3"
+  - julia -e 'if Sys.islinux() && string(VERSION)[1:3] == "1.4"
```
dpo: Could we just say "1" here?

abelsiqueira: No, because `VERSION` will be full with `MAJOR.MINOR.PATCH`. `string(VERSION)[1]` will match all. I could drop `string(VERSION)` and use only `Sys.islinux()`. It will slow down only a bit, but since OSX and Windows take the most time, it won't be noticeable.

dpo: Whatever's best. I don't have a preference. It's just that we have to remember to change that version number every time.

dpo:
Have to figure out what's going on with FreeBSD.

PR #221, opened 2020-04-11
fix CI scripts

dpo:


dpo:
Can't seem to be able to install gfortran on travisCI/freebsd. Skipping.

PR #224, opened 2020-05-09
package SIF collection as artifact

dpo:
Use Julia 1.3's artifact feature to download and install the SIF collection once and for all. According to the docs, the artifact will be shared by several installations of CUTEst.jl (say, in several environments), and so will be downloaded only once.

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -1,0 +1,6 @@
+[sifcollection]
+git-tree-sha1 = "a7ea0d0aaf29a39ca0fe75588fc077cdd5b5ed54"
+
+    [[sifcollection.download]]
+    url = "https://bitbucket.org/optrove/sif/get/99c5b38e7d03.tar.gz"
+    sha256 = "33b29e46288c6f019f4ee58e0df6b2797bc2642419b2c181edc574c8db8f6f34"
--- a/Project.toml
+++ b/Project.toml
@@ -9,6 +9,7 @@ JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NLPModels = "a4795742-8479-5a88-8948-cc11e1c8c1a6"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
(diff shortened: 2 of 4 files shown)


dpo:
I like the result but this was incredibly frustrating to get working.

dpo:
I agree that it would be good to fetch the collection first thing, but what if the user already has the collection and has set MASTSIF?

abelsiqueira:
Inside `__init__`, if the user doesn't have `MASTSIF` set, we define a default one. Instead, we could fetch.

Also, just noticed an error, `ENV["MASTSIF"]` is lost after closing `julia`, even after `fetch_sif_problems()`. So the default `MASTSIF` should at least point to the same artifact place.

dpo:
I set it to `CUTEst-2.0.3/sif` by default in case the user doesn't want to download the whole thing. Should we just go ahead and download?

Not sure I understand the error. Every time I restart Julia, MASTSIF points to `CUTEst-2.0.3/sif`, and if I call `fetch_sif_problems()`, it points to the whole collection (without downloading again).

abelsiqueira:
But then I have to call `fetch_sif_problems()` every time, which is counter-intuitive. I think we should download by default if there is no `MASTSIF`. The problems in `CUTEst-2.0.3/sif` are only useful for testing.

dpo:
I agree, but if the user has already set `MASTSIF`, then how do we change `MASTSIF` (after `__init__`) to point to the full collection?

abelsiqueira:
Ask the user to unset `MASTSIF`. If the user has a working `MASTSIF`, we won't need to use the artifact. If he has no `MASTSIF`, he will use the artifact. If he has a `MASTSIF` that doesn't work, he should unset it.
I'm looking into what is more common, and I think that most people will use CUTEst.jl without any CUTEst installation because they're using only Julia. In second most used, I think they'll be comparing a non-julia solver with a Julia solver, in which case he'll have both, so either MASTSIF will work.

Just noticed: do we have an `export_cutest_env()` so the user can install CUTEst.jl and use it outside of Julia?

dpo:
I now point to the main MASTSIF unless the env var is already set. The method `set_mastsif()` points it to the main collection. The user can explicitly change the env var in Julia if they want to point to some other location, but we needed a mechanism to reset to the artifact.

> do we have an export_cutest_env() so the user can install CUTEst.jl and use it outside of Julia?

Let's open an issue.

abelsiqueira:
Thanks

PR #226, opened 2020-05-11
print CUTEst environment variables

### LINKED ISSUE #225: Return environment so CUTEst can be used outside julia
Something simple like printing
```
CUTEST=...
```
and giving instructions on where to put it should be enough.

dpo:
closes #225

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -15,7 +15,7 @@ import Libdl.dlsym
 # Only one problem can be interfaced at any given time.
 global cutest_instances = 0
 
-export CUTEstModel, sifdecoder, set_mastsif
+export CUTEstModel, sifdecoder, set_mastsif, export_cutest_env
 
 mutable struct CUTEstModel <: AbstractNLPModel{Float64, Vector{Float64}}
   meta::NLPModelMeta{Float64, Vector{Float64}}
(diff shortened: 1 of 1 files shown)


dpo:
Doesn't work out of the box on macOS unfortunately, even after setting `DYLD_FALLBACK_LIBRARY_PATH`:
```
$ sifdecoder ROSENBR
dyld: Library not loaded: @rpath/libgfortran.5.dylib
  Referenced from: /Users/dpo/.julia/artifacts/83e267cb7d8652c72b9dac8674d69220810103c2/libexec/SIFDecode-2.0.3/objects/mac64.
osx.gfo/double/run_sifdecode
  Reason: image not found
/Users/dpo/.julia/artifacts/83e267cb7d8652c72b9dac8674d69220810103c2/libexec/SIFDecode-2.0.3/bin/sifdecoder: line 459: 48619 A
bort trap: 6           $DECODER
 ERROR: Error exit from decoding stage. terminating execution.
```

dpo:
The updated version works for me on macOS.

@abelsiqueira Does it work for you on Linux?

dpo:
There's still a caveat that running `runcutest` tries to update the CUTEst library, which results in a lot of "permission denied" messages being printed. It's inconsequential, but annoying.

abelsiqueira:
Unfortunately doesn't work for me either:
```
/home/abel/.julia/artifacts/c990de2425d3a17849e2f6bbe47decf0f81036d8/libexec/SIFDecode-2.0.3/objects/pc64.lnx.gfo/double/run_sifdecode: error while loading shared libraries: libgfortran.so.4: cannot open shared object file: No such file or directory
```

dpo:
Does it work if you define this bash function: https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/226/files#diff-4b9f14fdac93f5150aa5c070ec70fc63R129

abelsiqueira:
Almost worked, needs a `;` at the end: `... bash -c "..."; }`.
With that change, `run_sifdecoder ROSENBR` worked.

dpo:
Should be good now.

abelsiqueira:
We need a similar one for `runcutest`. I tried adding a
```
run_cutest() { ... source $CUTEST/bin/runcutest $*"; }
```
but it failed when I ran `run_cutest -p genc -D ROSENBR` with a bunch of these:
```
ar: /home/abel/packages/julias/julia-1.4-latest/bin/../lib/julia/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /usr/lib/libLLVM-9.so)
```
The `sifdecoder` part ran correctly.

PS. The `$*` instead of `$@` was necessary to parse the arguments correctly.

dpo:
Does this work though
```
run_sifdecoder ROSENBR
run_cutest -p genc
```

abelsiqueira:
No, it has the same error. The final line is `Error opening file OUTSDIF.d.`, which I guess it's because the compilation failed.

dpo:
Hm. It works for me on macOS. That seems to indicate an issue with the JLL... :-(

tmigot:
Thanks, I just noticed too.
I rebased this just in case, if this is still of interest ?

dpo:
It is, but what's a bit weird is that `runcutest` modifies the artifacts. It works, but it's not ideal. I should modify CUTEst itself so it only modifies files in `~/.cutest` (say), and leaves the compiled object files alone.

amontoison:
Can I close this PR?
All the environment variables have been removed because they not needed anymore with the new artifact `SIFDecode_jll.jl`.

PR #234, opened 2020-08-25
fix docstring to prevent italics

dpo:
Currently, the underscores are interpreted as begin/end italics, so it's not obvious that the keyword args have underscores.


abelsiqueira:
Many thanks. Travis failed randomly as usual, so I'm ignoring.

PR #240, opened 2020-10-29
Remove fetch_sif_problems from README

abelsiqueira:


README.md
```diff
@@ -32,14 +32,6 @@ The following command installs the CUTEst binaries and Julia interface:
 pkg> add CUTEst
 ````
 
-If you already have a collection of SIF problems that you wish to use, you can simply set the `MASTSIF` environment variable to point to their location.
-If not, the command
```
dpo, line 36: This part is still applicable: https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/master/src/CUTEst.jl#L67

If MASTSIF isn't already set, it will be set to the artifact location.

dpo (approved):
Thanks!

PR #250, opened 2021-03-11
Update to NLPModels 0.14 + NLPModelsTest

abelsiqueira:
Related to the refactoring of NLPModels.jl

cf @dpo @tmigot


dpo:
Error on macOS:
```
Error: gfortran is not installed. Please install it and try again.
```

PR #260, opened 2021-06-12
Update to NLPModels 0.15

abelsiqueira:
CUTEst only works with Float64, so the struct is specialized.
I don't know if it works with GPU. If it does, I can change to `S`.


dpo (approved):
Looks good. Thanks!

dpo:
PS: I don’t think anything here will work on gpu. We’ll have to convert vectors back and forth all the time.

PR #266, opened 2021-07-04
Bump to NLPModels 0.16 and 0.17 and update Version to 0.12.1

abelsiqueira:
No src files were changed, only the tests, so the version only has a minor bump


dpo (changes requested):
Thanks. Could you also update the tutorial as in #258, or update #258 directly?

abelsiqueira:
Done. Sorry, forgot about #258.

dpo (approved):

PR #269, opened 2021-08-28
Add support for view inputs

sshin23:
Hey JSO devs, we've been using CUTEst.jl for benchmarking our solver [MadNLP.jl](https://github.com/sshin23/MadNLP.jl), and we recently found that JSO's current julia interface does not efficiently handle view inputs. Please consider merging this PR.

This PR adds better support for view inputs for CUTEst.jl's julia interface.

The current CUTEst.jl julia interface converts any `AbstractVector` input to `Vector{Float64}` to make it compatible with subsequent ccalls, and this process accompanies memory allocations. But when the AbstractVector is a stride-one array (which typically comes from a view of a `Vector` over a `UnitRange`), this can be avoided by directly passing it to C functions. 

I have added the definition for `StrideOneVector{T}`, and replaced the type specifications in the functions from `::Vector{Float64}` to `::StrideOneVector{Float64}`. This change allows directly using stride-one vectors as inputs to the C functions instead of allocating more memory. I have also added a test for `StrideOneVector{T}` in `test_core.jl`.

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -105,6 +105,12 @@ macro cutest_error()  # Handle nonzero exit codes.
   esc(:(io_err[1] > 0 && throw(CUTEstException(io_err[1]))))
 end
 
+# to allow view inputs with stride one
+StrideOneVector{T}=Union{
+    Vector{T},
+    SubArray{T,1,Array{T,1},Tuple{UnitRange{U}},true} where {U<:Integer}
+}
+
 include("core_interface.jl")
 include("julia_interface.jl")
 include("classification.jl")
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -106,7 +106,7 @@ macro cutest_error()  # Handle nonzero exit codes.
 end
 
 # to allow view inputs with stride one
-StrideOneVector{T}=Union{
+StrideOneVector{T} = Union{
     Vector{T},
     SubArray{T,1,Array{T,1},Tuple{UnitRange{U}},true} where {U<:Integer}
 }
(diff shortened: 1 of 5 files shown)


src/CUTEst.jl
```diff
@@ -105,6 +105,12 @@ macro cutest_error()  # Handle nonzero exit codes.
   esc(:(io_err[1] > 0 && throw(CUTEstException(io_err[1]))))
 end
 
+# to allow view inputs with stride one
+StrideOneVector{T}=Union{
```
dpo, line 109:
```suggestion
StrideOneVector{T} = Union{
```

src/CUTEst.jl
```diff
@@ -105,6 +105,12 @@ macro cutest_error()  # Handle nonzero exit codes.
...
+    Vector{T},
```
dpo, line 110:
```suggestion
  Vector{T},
```

src/CUTEst.jl
```diff
@@ -105,6 +105,12 @@ macro cutest_error()  # Handle nonzero exit codes.
...
+    SubArray{T,1,Array{T,1},Tuple{UnitRange{U}},true} where {U<:Integer}
```
dpo, line 111:
```suggestion
  SubArray{T, 1, Vector{T}, Tuple{UnitRange{U}}, true} where {U <: Integer}
```

dpo (changes requested):
Many thanks for this. Just a few rather unsubstantial comments for now.

abelsiqueira (approved):
Thanks for the PR, very good stuff!

dpo:
thank you!

sshin23:
thanks for the merge!

Issue #270, opened 2021-08-29
`jac_coord!` memory allocation


sshin23:
The current implementation of `jac_coord!` internally creates `c = Vector{Float64}(undef, nlp.meta.ncon)`, which allocates memory. I was wondering if we can create a workspace array in `CUTEstModel` and reuse it whenever `jac_coord!` is called to avoid this allocation.

dpo:
That sounds like a good idea provided it's documented properly.

dpo:
fixed in #271

PR #271, opened 2021-08-30
Fix for jac_coord! allocation issue

sshin23:
This PR implements the changes suggested in #270.

To avoid memory allocation in `NLPModels.jac_coord!` and `NLPModels.obj`, a workspace array is added in `CUTEstModel`. The workspace array is used within `NLPModels.jac_coord!` and `NLPModels.obj` to avoid creating new arrays.

I believe no change in the test is necessary for this PR.


dpo:
thanks!

PR #281, opened 2022-07-10
Fix unit tests for problem hs14

tmigot:
Since https://github.com/JuliaSmoothOptimizers/NLPModelsTest.jl/pull/50/files there nonlinear function in NLPModelsTest.jl and CUTEst.jl now differs. So, I suggest to test with the function (defined in this PR) `compare_cons`.

The unit tests then break because of the linear API, see https://github.com/JuliaSmoothOptimizers/CUTEst.jl/tree/add-linear-API (with NLPModelsTest 0.8.1), coming in a following PR


abelsiqueira:
Thanks for the PR

PR #283, opened 2022-07-11
Add meta for linear API

tmigot:
Add `lin_nnzj` and `nln_nnzj` in the meta.
There is no specific API to get the linear jacobian in CUTEst, so I used `nnzj` as an upper bound.

The tests break in this branch, the working one is here: https://github.com/JuliaSmoothOptimizers/CUTEst.jl/tree/add-linear-API

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -7,4 +7,4 @@ NLPModels = "a4795742-8479-5a88-8948-cc11e1c8c1a6"
 [compat]
 Documenter = "~0.26"
 Krylov = "0.7.0"
-NLPModels = "0.17.0"
+NLPModels = "0.18"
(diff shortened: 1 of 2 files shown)


dpo:
Nick Gould's recommendation was to evaluate the gradient of each linear constraint individually.

tmigot:
> Nick Gould's recommendation was to evaluate the gradient of each linear constraint individually.

That's what I planned for the jacobian structure/values https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/54401c7a4f230d5ac2de74f4c0f0fde5f720c597/src/julia_interface.jl#L373 , but I wanted to skip evaluating it for the number of nonzeros.

PR #284, opened 2022-07-11
Add linear API

tmigot:


src/julia_interface.jl
```diff
@@ -272,6 +272,44 @@ function NLPModels.cons!(nlp::CUTEstModel, x::AbstractVector, c::AbstractVector)
...
 
+function NLPModels.cons_lin!(nlp::CUTEstModel, x::AbstractVector, c::AbstractVector)
+  _cx = Vector{Float64}(undef, nlp.meta.nlin)
+  cons_lin!(nlp, x, _cx)
+  c .= _cx
+  return c
+end
+
+function NLPModels.cons_lin!(nlp::CUTEstModel, x::AbstractVector, c::StrideOneVector)
+  st = view(Cint[0; 0], 1:1)
```
dpo, line 283: why isn't `Cint[0]` sufficient here?

src/julia_interface.jl
```diff
@@ -332,12 +370,110 @@ function NLPModels.jac_structure!(
...
+  rows::AbstractVector{<:Integer},
+  cols::AbstractVector{<:Integer},
+)
+  st = view(Cint[0; 0], 1:1)
+  nvar = view(Cint[nlp.meta.nvar], 1:1)
+  ci = view([0.0], 1:1)
+  nnzj = view(Cint[0], 1:1)
+  lj = view(Cint[nlp.meta.nvar], 1:1)
+  Jval = @view(Array{Cdouble}(undef, nlp.meta.nvar + 1)[2:end])
+  Jvar = @view(Array{Cint}(undef, nlp.meta.nvar + 1)[2:end])
```
dpo, line 384: This looks incredibly inefficient. I think we would be better off calling `jac_structure!()` and separating out ourselves the linear and the nonlinear part, and storing them inside the nlp. What do you think?

dpo: Or at the very least, Jval and Jvar should be preallocated and stored inside the nlp.

src/julia_interface.jl
```diff
@@ -332,12 +370,110 @@ function NLPModels.jac_structure!(
...
 end
 
+function NLPModels.jac_lin_coord!(nlp::CUTEstModel, x::AbstractVector, vals::AbstractVector)
+  st = view(Cint[0; 0], 1:1)
+  nvar = view(Cint[nlp.meta.nvar], 1:1)
+  ci = view([0.0], 1:1)
+  nnzj = view(Cint[0], 1:1)
+  lj = view(Cint[nlp.meta.nvar], 1:1)
+  Jval = @view(Array{Cdouble}(undef, nlp.meta.nvar + 1)[2:end])
+  Jvar = @view(Array{Cint}(undef, nlp.meta.nvar + 1)[2:end])
```
dpo, line 440: same comments here.

src/julia_interface.jl
```diff
@@ -400,6 +536,32 @@ function NLPModels.jprod!(
...
+end
+
+function NLPModels.jprod_lin!(
+  nlp::CUTEstModel,
+  x::AbstractVector,
+  v::AbstractVector,
+  jv::AbstractVector,
+)
+  jvc = zeros(nlp.meta.ncon)
+  jprod!(nlp, Vector{Float64}(x), Vector{Float64}(v), jvc)
```
dpo, line 559: since these constraints are linear, the Jacobian is constant and could be stored. Then we could compute a product efficiently.

dpo (changes requested):

tmigot:
This is ready for another review. Thanks for the comments, I think I fixed most of them and tried to improve.

dpo (commented):
Looking good, thank you! The worry is that all those loops are slow, but currently, there's no other way.

src/julia_interface.jl
```diff
@@ -332,12 +369,88 @@ function NLPModels.jac_structure!(
...
+)
+  jac_structure!(nlp)
+  k = 1
+  for i in findall(j -> j in nlp.meta.nln, nlp.jrows)
+    rows[k] = nlp.jrows[i] - count(x < nlp.jrows[i] for x in nlp.meta.lin)
+    cols[k] = nlp.jcols[i]
+    k += 1
+  end
+  rows[k:end] .= zero(Int32)
+  cols[k:end] .= zero(Int32)
```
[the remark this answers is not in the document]
tmigot: For `nnzj_nln`, I used `min(nvar * (ncon - nlin), nnzj)` (see https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/e89575dc637af55ef78a09e804638cfaadbd79a0/src/CUTEst.jl#L349 from a previous PR).

The reason is that CUTEst only gives `nnzj` and I didn't want to evaluate the jacobian when first instantiating the model.

src/julia_interface.jl
```diff
@@ -272,6 +272,40 @@ function NLPModels.cons!(nlp::CUTEstModel, x::AbstractVector, c::AbstractVector)
...
+
+function NLPModels.cons_lin!(nlp::CUTEstModel, x::AbstractVector, c::StrideOneVector)
+  k = 1
+  for j in nlp.meta.lin
+    cifn(Cint[0], Cint[nlp.meta.nvar], Cint[j], x, view(c, k:k))
+    k += 1
+  end
+  nlp.counters.neval_cons_lin += 1
+  return c
+end
```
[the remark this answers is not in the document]
tmigot: From Nick Gould's advice, I understood that CUTEst doesn't redo computations for the linear constraints, so it was ok.

Still, I get your idea. Do you suggest we also compute the linear constraint within the function https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/e89575dc637af55ef78a09e804638cfaadbd79a0/src/julia_interface.jl#L372 so that `cons_lin` just do the matrix-vector product + vector ?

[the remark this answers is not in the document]
dpo: Yes, but maybe just open an issue about it. That could be done later.

src/julia_interface.jl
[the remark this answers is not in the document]
dpo: Ok thanks!

dpo (approved):

PR #285, opened 2022-08-01
use convert instead of Vector(x)

dpo:
If x is already a Vector{Float64} and y = convert(Vector{Float64}, x),
then x == y and x === y, which isn't true if y = Vector{Float64}(x).


tmigot (approved):
Thanks for updating!

PR #290, opened 2022-08-08
bump to Julia 1.6

dpo:


tmigot (changes requested):
Thanks! The CompatHelper.yml and .cirrus files also need to be updated for this.

dpo:
Thanks. Done.

PR #294, opened 2022-08-11
add `lencheck` everywhere

tmigot:
This allows to replace things like `hv[1:(nlp.meta.nvar)] .= hvc` by `hv .= hvc`


dpo:
I agree. The user can always pass in views.

dpo (approved):

PR #299, opened 2022-09-26
Release 0.13.1

amontoison:


amontoison:
The artifact CUTEST_jll is very old and I will regenerate a new one with Yggdrasil.
By default we should lock the version of the artifact is the new version breaks things.

tmigot (approved):
Ok, thanks!

PR #314, opened 2023-10-01
Added `increment!` and `decrement!` functions.

### LINKED ISSUE #292: Use `increment!` and `decrement!` functions
instead of doing this https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/7ea04f48b9d972fb8ce37e160358a0d103186474/src/julia_interface.jl#L43

use the `increment!` function like here
https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/d4c48d2fd7d599aee83b97624d3299b3fcd98c1f/src/nlp.jl#L537

The function is defined here: https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/main/src/nlp/counters.jl#L59

Jay-sanjay:
Fixes #292 
updated julia_interface.jl

### DIFF
--- a/src/julia_interface.jl
+++ b/src/julia_interface.jl
@@ -2,13 +2,27 @@ export cons_coord, cons_coord!, consjac
 
 using NLPModels, SparseArrays
 
-function increment!(x::Number, n::Number)
-  x += n
-  return x
+"""
+    increment!(nlp, s)
+
+Increment counter `s` of problem `nlp`.
+"""
+@inline function increment!(nlp::AbstractNLPModel, s::Symbol)
+  increment!(nlp, Val(s))
 end
-function decrement!(x::Number, n::Number)
-  x -= n
-  return x
+
+for fun in fieldnames(Counters)
+  @eval increment!(nlp::AbstractNLPModel, ::Val{$(Meta.quot(fun))}) = nlp.counters.$fun += 1
+end
+
+"""
+    decrement!(nlp, s)
+
+Decrement counter `s` of problem `nlp`.
+"""
+
+function decrement!(nlp::AbstractNLPModel, s::Symbol)
+  setproperty!(nlp.counters, s, getproperty(nlp.counters, s) - 1)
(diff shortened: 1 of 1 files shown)


abelsiqueira:
Hi @Jay-sanjay, thanks for the PR. Unfortunately, that is not the `increment!` that we wanted - our fault for not being clear on the issue. Check it here: https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl/blob/main/src/nlp.jl#L537

Jay-sanjay:
@abelsiqueira  sir do check is it the way the `increment!()` must be done , if yes I will update it for all other quantities too , that I did earlier

src/julia_interface.jl
```diff
@@ -2,6 +2,29 @@ export cons_coord, cons_coord!, consjac
...
+"""
+    decrement!(nlp, s)
+
+Decrement counter `s` of problem `nlp`.
+"""
+
+function decrement!(nlp::AbstractNLPModel, s::Symbol)
+  setproperty!(nlp.counters, s, getproperty(nlp.counters, s) - 1)
+end
+
```
abelsiqueira, line 27: These shouldn't be necessary since they are already defined in NLPModels.jl. Just remove them.

src/julia_interface.jl
```diff
@@ -43,7 +66,7 @@ function NLPModels.objcons!(
       f,
     )
   end
-  nlp.counters.neval_obj += 1
+  increment!(nlp, :neval_obj)
```
abelsiqueira, line 69: Yes, this is the correct usage.

abelsiqueira (changes requested):
Hi @Jay-sanjay, thanks. That is the correct usage, but you don't have to define the functions again since they are already present in the NLPModels.jl package.

abelsiqueira (approved):
Thanks @Jay-sanjay, this looks great.

PR #319, opened 2024-01-09
Add an option to use our own version of CUTEst

amontoison:
@AndrewLister-STFC

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -2,7 +2,7 @@ task:
   matrix:
     - name: FreeBSD
       freebsd_instance:
-        image_family: freebsd-13-1
+        image_family: freebsd-13-2
       env:
         matrix:
           - JULIA_VERSION: 1.6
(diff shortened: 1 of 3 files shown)


dpo:
I’m not a big fan of this. The Julia interface could be out of sync with the version of CUTEst that the user has installed. Instead, I would very much prefer that we cook up a way to use the CUTEst JLL from outside Julia.

amontoison:
This backdoor is not documented.
It's a way to test new modifications or a new version of a package before that we precompile them with Yggdrasil.
It's what we do for HSL.jl, GALAHAD.jl, MUMPS.jl, QRMumps.jl, etc...

amontoison:
I also try to recompile CUTEst with Yggdrasil (https://github.com/JuliaPackaging/Yggdrasil/pull/7924) for Andrew.
He sent me an email for that.
He would like to use the single precision version of CUTEst that Nick recently updated with the Julia interface CUTEst.jl.

dpo:
Then we should at least display `@warn` messages alerting the user that this usage is not supported and that they’re using an external CUTEst library at their own risk.

PR #320, opened 2024-01-09
Upgrade CUTEst_jll

amontoison:


amontoison:
Hum, it seems that Nick hardcode the name of the Fortran compiler when we compile  the binary `sifdecoder` of the package `SIFDecode`.
The issue is that we define the Fortran compiler with an environment variable `FC`  to handle multiple platforms in the cross-compiler BinaryBuilder.
The culprit is [here](https://github.com/ralna/ARCHDefs/blob/master/compiler.binarybuilder.bb.fc#L3) in ARCHDefs.

```shell
/home/alexis/.julia/artifacts/51305f685c9b2d4845a565c499054718552d1653/bin/sifdecoder: ligne 242: FC : commande introuvable
```
cc @nimgould

nimgould:
These sifdecode/cutest scripts are intended to be used with the archdefs files. A default value FORTRAN="f95" is provided, but this is overwritten by the appropriate architecture file using eval "`cat $SIFDECODE/bin/sys/$ARCH`". None of this is relevant to anything other than fortran testing using CUTEst, and really it isn't necessary for C/python/jula/matlab/octave use (each of these have their own avenues to using SIF files). The optrove (sifdecode/cutest/galahad) scripts use the variables (e.g.) 
[pc64.lnx.gfo.txt](https://github.com/JuliaSmoothOptimizers/CUTEst.jl/files/13870864/pc64.lnx.gfo.txt)
These names are in GALAHAD documentation going back to the 1990s, so really shouldn't be changed

tmigot:
@amontoison can we close this?

amontoison:
No, we would like to use shared libraries but the current version are incomplete.
We need to trick the linker such that we can generate in `.so` or `.dylib` but we can't with a `.dll`.
I solved the issue by doing a demuxer of symbols but we need the release 2.2.1 of CUTEst_jll.jl for that.

PR #327, opened 2024-05-17
Bump to NLPModels 0.21

tmigot:


tmigot:
Need https://github.com/JuliaSmoothOptimizers/NLPModels.jl/pull/463

tmigot:
@amontoison @dpo @abelsiqueira  Not sure I have seen this error before https://github.com/JuliaSmoothOptimizers/CUTEst.jl/actions/runs/9226211936/job/25385471817?pr=327#step:7:162 any idea?

dpo:
Sorry, I missed this. Is the error occurring because of the NLPModels upgrade or because of an upgrade of CUTEst/SifDecode?

src/julia_interface.jl
```diff
@@ -44,6 +43,7 @@ function NLPModels.objcons!(
     )
   end
   increment!(nlp, :neval_obj)
+  increment!(nlp, :neval_cons)
```
dpo, line 46: Well, isn't this the source of the error? You're incrementing `nevals_cons` even on unconstrained problems.

More generally, this pull request is about bumping the NLPModels version, but it does other things. I suggest removing the other changes so we can isolate the problem (if any).

dpo (changes requested):

src/julia_interface.jl
tmigot: This is part of the update to NLPModels 0.21. This was necessary to pass unit tests from NLPModelsTest that come along the new release.

Essentially, the idea is that we need to be consistent when calling cons! (and objcons!) on unconstrained problems.
I opened a new PR that updates link to better isolate the issue with the CI on macOS https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/328

dpo: But what sense is there to incrementing `neval_cons` on an unconstrained problem?

amontoison:
@tmigot Is it possible to support NLPModels.jl v0.21 now?

PR #328, opened 2024-05-30
Update link to citation file

tmigot:


tmigot:
@dpo a PR that better isolate the issue on macOS

dpo:
@tmigot This is a different error. Here the issue is that you're building on Mac x64, but CUTEst_jll.jl installs the aarch64 shared libraries (probably because of the gfortran version). I noticed the same in my recent SolverBenchmark PR and updated to workflow: see lines 18-26 of https://github.com/JuliaSmoothOptimizers/SolverBenchmark.jl/blob/main/.github/workflows/examples.yml#L18

amontoison:
You need `macos-13` if you want to the architecture `x64`, `macos-latest` is `macos-14` and only available on recent Mac with the architecture `arm64`.

PR #332, opened 2024-08-14
Generate Julia wrappers with Clang.jl

amontoison:
- Generate Julia wrappers for single and double precision of CUTEst

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ deps/*
 
 docs/build
 docs/site
-dos/Manifest.toml
+docs/Manifest.toml
 
 gen/Manifest.toml
 
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ deps/*
 docs/build
 docs/site
 dos/Manifest.toml
+
+gen/Manifest.toml
+
--- a/gen/wrapper.jl
+++ b/gen/wrapper.jl
@@ -1,4 +1,4 @@
-# Script to parse qr_mumps headers and generate Julia wrappers.
+# Script to parse CUTEst headers and generate Julia wrappers.
 using CUTEst_jll
 using Clang
 using Clang.Generators
--- a/gen/wrapper.jl
+++ b/gen/wrapper.jl
@@ -1,0 +1,44 @@
+# Script to parse qr_mumps headers and generate Julia wrappers.
+using CUTEst_jll
+using Clang
+using Clang.Generators
+using JuliaFormatter
+
+function main()
+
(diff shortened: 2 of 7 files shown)


gen/wrapper.jl
```diff
@@ -0,0 +1,44 @@
+# Script to parse qr_mumps headers and generate Julia wrappers.
```
tmigot, line 1:
```suggestion
# Script to parse CUTEst headers and generate Julia wrappers.
```

.gitignore
```diff
@@ -9,3 +9,6 @@ deps/*
 docs/build
 docs/site
 dos/Manifest.toml
```
tmigot, line 11:
```suggestion
docs/Manifest.toml
```

src/CUTEst.jl
```diff
@@ -215,8 +218,8 @@ function sifdecoder(
       end
       run(`mv OUTSDIF.d $outsdif`)
       delete_temp_files(suffix)
-      global cutest_lib =
-        Libdl.dlopen(libname, Libdl.RTLD_NOW | Libdl.RTLD_DEEPBIND | Libdl.RTLD_GLOBAL)
+      global cutest_lib = Libdl.dlopen(libname, Libdl.RTLD_NOW | Libdl.RTLD_DEEPBIND | Libdl.RTLD_GLOBAL)
```
tmigot, line 221: This will be modified again by the formatter, no?

src/libcutest.jl
```diff
@@ -0,0 +1,1008 @@
+function CUTEst_malloc(object, length, s)
```
tmigot, line 1: The follow-up PR is to use this interface in the code, right? I guess most of https://github.com/JuliaSmoothOptimizers/CUTEst.jl/blob/main/src/core_interface.jl will be updated then, and you can reuse the docstrings

tmigot (changes requested):

src/CUTEst.jl
amontoison: Yes, but I modified the name of the variable before that I reverted it.

src/libcutest.jl
amontoison: Yes, that's the plan.
I already did the modification on my laptop.
It will open a PR for you.
It should be easy to implement CUTEstModel in single and quadruple precision now.

src/CUTEst.jl
amontoison: The auto formater will probably fix other things.

amontoison:
Thanks for the review @tmigot!

PR #333, opened 2024-08-14
Use the Julia wrappers in CUTEst.jl -- Part I

amontoison:
- The function `set_mastsif` is now in `sifdecoder.jl`
- All code related to `CUTEstModel` is in a new file `model.jl`.
- Use the Julia wrappers for `fortran_open_` and `fortran_close_`.


tmigot:
Tests are not passing

amontoison:
I know...

src/model.jl
```diff
@@ -0,0 +1,283 @@
...
+    else
+      msg = "unknown error"
+    end
+    return new(info, msg)
+  end
+end
+
+CUTEstException(info::Integer) = CUTEstException(convert(Int32, info))
+
+function cutest_error(status::Cint)  # Handle nonzero exit codes.
```
tmigot, line 270: Why change the macro?

src/model.jl
```diff
@@ -0,0 +1,283 @@
+mutable struct CUTEstModel <: AbstractNLPModel{Float64, Vector{Float64}}
```
tmigot, line 1: Did things change in this file? Since you move+modify it's really hard to see

gen/wrapper.jl
```diff
@@ -32,7 +32,6 @@ function main()
   code = replace(code, "Ptr{ip_}" => "Ptr{Cint}")
   code = replace(code, "Ptr{ipc_}" => "Ptr{Cint}")
   write(path, code)
-
```
tmigot, line 35: :)

src/model.jl
amontoison: Because the macro was not applied to a parameter directly. It was checking a variable `io_err` that is defined in the same scoop of variable but outside the macro. It can break for multiple reasons (scoop, name of the variable, etc...)

src/model.jl
amontoison: I updated the calls to `fortran_close_` and `fortran_open_` to just use the new wrappers.
I also changed vectors of one coefficient into references.
It requires less memory like this.

gen/wrapper.jl
amontoison: I wanted it without the space ;)

PR #338, opened 2024-08-15
Update Artifacts.toml

### LINKED ISSUE #337: Update Artifacts.toml
Repository: https://bitbucket.org/optrove/workspace/repositories/

The new release of the `sifcollection` is available at:
https://bitbucket.org/optrove/sif/get/229e00b81891789fe29f099d1844565ef0ac14d3.tar.gz

amontoison:
Related to #337 
- Use the last release of the `sifcollection` added yesterday by Nick.
- Add two new artifacts for LP and QP problems (`netlib-lp` and `maros-meszaros`).
- Add an attribute `lazy=true` to only download an artifact if we need it.

For example, we should not download any artifact if the user defines the environment variable `MASTSIF`.

Note that the new standalone decoder available in `SIFDecode_jll.jl` and used since `CUTEst_jll.jl` v0.13.3 is tested on all these problems: https://github.com/ralna/SIFDecode/pull/20
I tested it with `single`, `double` and `quadruple` precision.

We could update the function `set_matsif` to provide an option to switch between `sifcollection`, `maros-meszaros` and `netlib-lp` sets in a future PR.

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -1,6 +1,23 @@
 [sifcollection]
-git-tree-sha1 = "a7ea0d0aaf29a39ca0fe75588fc077cdd5b5ed54"
+git-tree-sha1 = "a39ce7e15e83de78fee4847fa67c5f63e3f0c3c5"
+lazy = true
 
     [[sifcollection.download]]
-    url = "https://bitbucket.org/optrove/sif/get/99c5b38e7d03.tar.gz"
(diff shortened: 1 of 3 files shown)


tmigot (approved):
Thanks!

amontoison:
@tmigot I added an optional argument to `set_mastsif` such that we can use another set of SIF problems.

tmigot (approved):
Still good for me

PR #339, opened 2024-08-16
Add a standalone sifdecoder function

### LINKED ISSUE #335: Update the function sifdecoder
We should add a standalone function `sifdecoder` that takes as inputs:
- the path of the SIF file
- the path of the folder where .f and .d should be generated
- additional options to decode the SIF file

Right now we decode and compile a shared library for a problem in the same time.
We should split these steps in two functions.

amontoison:
close #335 

- Add a new file `sifdecoder.jl`
- Split the previous function `sifdecoder` into two new functions `sifdecoder` and `build_libsif`
- Add CI on Windows
- Test the new function `sifdecoder` on all platforms
- Add an option `libsif_folder` to specify where we want to decode the SIF problems and build the shared libraries

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         version: ['1.6', '1']
-        os: [ubuntu-latest, macos-13]
+        os: [ubuntu-latest, windows-latest, macos-13]
         arch: [x64]
         allow_failure: [false]
         include:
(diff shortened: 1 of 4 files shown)


amontoison:
@jfowkes The binary `sifdecoder_standalone` is working on Windows!
It could be relevant for `PyCUTEst`.
https://github.com/JuliaSmoothOptimizers/CUTEst.jl/actions/runs/10426477345/job/28879483562?pr=339#step:7:151

tmigot (approved):
That's great ! Thanks @amontoison

PR #344, opened 2024-08-17
Use the Julia wrappers in CUTEst.jl -- Part III

amontoison:
- use the Julia wrappers in `libcutest.jl` to easily update `CUTEst_jll.jl` in the future and support various precision.
- Update functions in `core_interface.jl` to use the C version `*_cint_*` instead of the F77 version because the type `logical` can have different sizes depending on the Fortran compiler.

The following routines are concerned:
- cutest_cint_csetup
- cutest_cint_cnoobj
- cutest_cint_uofg
- cutest_cint_ueh
- cutest_cint_ugreh
- cutest_cint_uhprod
- cutest_cint_ushprod
- cutest_cint_cofg
- cutest_cint_cofsg
- cutest_cint_ccfg
- cutest_cint_clfg
- cutest_cint_cgr
- cutest_cint_csgr
- cutest_cint_csgrp
- cutest_cint_csjp
- cutest_cint_ccfsg
- cutest_cint_ccifg
- cutest_cint_ccifsg
- cutest_cint_cgrdh
- cutest_cint_csgrsh
- cutest_cint_csgrshp
- cutest_cint_csgreh
- cutest_cint_chprod
- cutest_cint_chsprod
- cutest_cint_chcprod
- cutest_cint_cshcprod
- cutest_cint_chjprod
- cutest_cint_cjprod
- cutest_cint_csjprod
- cutest_cint_cchprods
- cutest_cint_cchprodsp
- cutest_cint_cohprods

PR #350, opened 2024-08-18
[CUTEst.jl] Support Windows

amontoison:
:crossed_fingers: 
Should be merged after #344 and #349 if it works one day...

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -28,12 +28,14 @@ const cutest_problems_path = joinpath(dirname(@__FILE__), "..", "deps", "files")
 isdir(cutest_problems_path) || mkpath(cutest_problems_path)
 
 function __init__()
-  if success(`bash -c "type gfortran"`)
-    @static Sys.isapple() ? (global libgfortran = []) :
-            (global libgfortran = [strip(read(`gfortran --print-file libgfortran.so`, String))])
-  else
-    @error "gfortran is not installed. Please install it and try again."
-    return
+  if !Sys.iswindows()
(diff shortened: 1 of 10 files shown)


amontoison:
Wow, it works :flushed: :astonished: :scream: 
cc @dpo

src/CUTEst.jl
```diff
@@ -28,12 +28,14 @@ const cutest_problems_path = joinpath(dirname(@__FILE__), "..", "deps", "files")
 isdir(cutest_problems_path) || mkpath(cutest_problems_path)
 
 function __init__()
```
tmigot, line 30:
```suggestion
function __init__()
```
Could you just add a comment around to say why we don't do anything for Windows here?

tmigot (changes requested):
That's really cool @amontoison, good job!
I am assuming this wouldn't work for anyone just testing directly on windows though, so we need a bit of documentation at least to warn them.

src/CUTEst.jl
amontoison: It's because we don't have `bash` on Windows. I will add a comment in a different PR.

amontoison:
> That's really cool @amontoison, good job! I am assuming this wouldn't work for anyone just testing directly on windows though, so we need a bit of documentation at least to warn them.

Yes, we require the user to have a Fortran compiler `gfortran` for now but I know how to install it as an artifact for users so it will be very user-friendly in the future. I will reuse the artifacts of `Artifacts.toml` from PackageCompiler.jl.

PR #352, opened 2024-08-18
Interface jth_hess_coord!

amontoison:
close #317 
Should be merged after https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/344, https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/349, https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/350 and #351

PR #353, opened 2024-08-18
Prepare CUTEst.jl for single and quadruple precision

amontoison:
Should be merged after https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/344, https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/349, https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/350, https://github.com/JuliaSmoothOptimizers/CUTEst.jl/pull/351 and #352

### DIFF
--- a/src/CUTEst.jl
+++ b/src/CUTEst.jl
@@ -1,8 +1,4 @@
-#See JuliaSmoothOptimizers/NLPModels.jl/issues/113
-__precompile__()
-
 # Using CUTEst from Julia.
-
 module CUTEst
 
 using CUTEst_jll
@@ -14,8 +10,12 @@ using NLPModels
 import Libdl.dlsym
 
 # Only one problem can be interfaced at any given time.
-global cutest_instances = 0
-global cutest_lib = C_NULL
+global cutest_instances_single = 0
+global cutest_instances_double = 0
+global cutest_instances_quadruple = 0
+global cutest_lib_single = C_NULL
+global cutest_lib_double = C_NULL
+global cutest_lib_quadruple = C_NULL
 
 export CUTEstModel, sifdecoder, build_libsif, set_mastsif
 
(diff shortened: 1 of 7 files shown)


src/CUTEst.jl
```diff
@@ -14,8 +10,12 @@ using NLPModels
...
 
 # Only one problem can be interfaced at any given time.
-global cutest_instances = 0
-global cutest_lib = C_NULL
+global cutest_instances_single = 0
+global cutest_instances_double = 0
+global cutest_instances_quadruple = 0
+global cutest_lib_single = C_NULL
+global cutest_lib_double = C_NULL
+global cutest_lib_quadruple = C_NULL
```
tmigot, line 18: Maybe we should at some point create a structure for this? That could be an issue with the tag `good first issue`.

src/model.jl
```diff
@@ -1,19 +1,19 @@
-mutable struct CUTEstModel <: AbstractNLPModel{Float64, Vector{Float64}}
-  meta::NLPModelMeta{Float64, Vector{Float64}}
+mutable struct CUTEstModel{T} <: AbstractNLPModel{T, Vector{T}}
```
tmigot, line 1: Really cool!

src/model.jl
```diff
@@ -226,7 +226,7 @@ function CUTEstModel(
   nnzj = nnzj[] |> Int
   nnzh = nnzh[] |> Int
 
-  work = Vector{Int32}(undef, ncon)
+  work = Vector{Float64}(undef, ncon)
```
tmigot, line 229: Could you add a one-line comment for this vector?

tmigot (changes requested):
Thanks @amontoison, looking good. Just some small comments

src/CUTEst.jl
amontoison: I added a SIF demuxer in the latest version of CUTEst such that we don't need anymore to update the global variables.
The issue is that package was not updated since long time we need many modifications if we want to easily upgrade it.

src/model.jl
amontoison: I think we should add two vectors in the structure (one of size `nvar` and another of size `ncon`). We need these kind of buffer / workspace for a lot of CUTEst routines.

PR #364, opened 2024-08-19
Update classf.json

### LINKED ISSUE #361: Regenerate a new classf.json
The SIF collection was updated recently, we should generate a new JSON file.
```julia
using CUTEst
CUTEst.create_class()
```
Note that it's quite long to generate and require a lot of storage...

amontoison:
close #361 
It seems that some SIF files were removed from the SIF collection but they are still in the database (`CLASSF.DB`):
- HIER133B
- HIER133C
- HIER163B
- HIER163C
- HIER163E
- TABLE4
- TABLE5

cc @nimgould

### DIFF
--- a/src/classf.json
+++ b/src/classf.json
@@ -4038,13 +4038,13 @@
     "has_internal_var": true,
     "derivative_order": 2,
     "variables": {
-      "bounded_below": 6905,
+      "bounded_below": 5,
       "number": 6907,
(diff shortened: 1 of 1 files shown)


nimgould:
Thank you, yes these examples were removed from the collection in August 2022 (see sif.updates)  but not from the database. Now fixed.

PR #365, opened 2024-08-20
Use an artifact on Windows for gfortran

amontoison:


amontoison:
@tmigot All users will be able to use CUTEst.jl on Windows now :smiley:

tmigot:
That's really cool! Thanks @amontoison

PR #366, opened 2024-08-20
Add single precision in core_interface.jl

amontoison:


tmigot (approved):
Looks good to me, it will be easier to debug the single precision with the unit tests to follow.

amontoison:
> Looks good to me, it will be easier to debug the single precision with the unit tests to follow.

We need a new version of `CUTEst_jll.jl` to use single and quadruple precision.
The issue is that I have an error in the garbage collector so I only allow double precision for now.
I suspect that the error is upstream.

PR #375, opened 2024-08-21
Improve obj and grad!

amontoison:


amontoison:
Memory leak in `cutest_cigr_`...
```shell
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x6c301825 -- .text at D:\a\CUTEst.jl\CUTEst.jl\deps\files\libHS6_double.DLL (unknown line)
in expression starting at D:\a\CUTEst.jl\CUTEst.jl\test\nlpmodelstest.jl:1
.text at D:\a\CUTEst.jl\CUTEst.jl\deps\files\libHS6_double.DLL (unknown line)
.text at D:\a\CUTEst.jl\CUTEst.jl\deps\files\libHS6_double.DLL (unknown line)
cutest_cigr_ at D:\a\CUTEst.jl\CUTEst.jl\deps\files\libHS6_double.DLL (unknown line)
cutest_cigr_ at D:\a\CUTEst.jl\CUTEst.jl\src\libcutest.jl:411
cigr at D:\a\CUTEst.jl\CUTEst.jl\src\core_interface.jl:3312 [inlined]
grad! at D:\a\CUTEst.jl\CUTEst.jl\src\julia_interface.jl:32
```

src/julia_interface.jl
```diff
@@ -2,10 +2,47 @@ export cons_coord, cons_coord!, consjac
 
 using NLPModels, SparseArrays
 
-function NLPModels.objcons(nlp::CUTEstModel{T}, x::AbstractVector) where {T}
```
dpo, line 5: I don't think you want to remove `objcons` because in CUTEst, it's more efficient to evaluate them together than separately. The default `objcons` will evaluate them separately.

src/julia_interface.jl
```diff
@@ -2,10 +2,47 @@ export cons_coord, cons_coord!, consjac
...
+  else
+    ufn(T, status, nvar, x, f)
+  end
+  increment!(nlp, :neval_obj)
+  return f[]
+end
+
+function NLPModels.obj(nlp::CUTEstModel{T}, x::AbstractVector) where T
+  @lencheck nlp.meta.nvar x
+  x_ = Vector{T}(x)
```
dpo, line 22: Maybe this could be preallocated in the problem structure?

src/julia_interface.jl
```diff
@@ -2,10 +2,47 @@ export cons_coord, cons_coord!, consjac
...
+    ugr(T, status, nvar, x, g)
+  end
+  increment!(nlp, :neval_grad)
+  return g
+end
+
+function NLPModels.grad!(nlp::CUTEstModel{T}, x::AbstractVector, g::AbstractVector) where T
+  @lencheck nlp.meta.nvar x g
+  x_ = Vector{T}(x)
+  g_ = Vector{T}(g)
```
dpo, line 43: Preallocate this too?

src/julia_interface.jl
amontoison: `objcons` is implemented by default in `NLPModels.jl`:
https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/main/src/nlp/api.jl#L114-L123

We just need to implement `objcons!`.

src/julia_interface.jl
amontoison: Yes, we could but it's the same issue with all routines.
I'm wondering if we should not remove the dispatch for input vectors of wrong type.
If the `CUTEstModel` is in double precision, the user should provide input arrays in double precision.
What do you think?

src/julia_interface.jl
dpo: Ok.

src/julia_interface.jl
dpo: Most of Julia dispatches in that way. It’s useful because we routinely call problem functions with views.

amontoison: Views are handled by the other method with `x::StrideOneVector{T}`.

This kind of method is only used if `x` is a `Vector{Float32}` or a `CuVector{Float64}` and nlp is a `CUTEstModel{Float64}`.
But should we handle that?

PR #378, opened 2024-08-23
Interface routines for quadruple precision

amontoison:
@oscardssmith @RalphAS 
Do you know what is the best way to provide a `Vector` of a `Ref` of `Float128` to a Fortran routine?
The Fortran routines probably needs a Vector of `Cfloat128`, am I right?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,6 +20,7 @@ Combinatorics = "1.0"
 DataStructures = "0.17, 0.18"
 JSON = "0.8, 0.9, 0.10, 0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17, 0.18, 0.19, 0.20, 0.21"
 NLPModels = "0.18, 0.19, 0.20"
+Quadmath = "0.5.10"
 SIFDecode_jll = "2.6.0"
 julia = "^1.6.0"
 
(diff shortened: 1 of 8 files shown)


RalphAS:
I've just used `Ptr{Float128}` in the signature list and `Array{Float128}` or `RefValue{Float128}` objects in the actual argument list of the `ccall`, and relied on the default conversion methods. But then I've only applied this to my own Fortran code which has `use iso_c_binding` and `real(kind=C_FLOAT128)`.

`Cfloat128` also seems to work for my library (changing only the Julia wrapper), with a simple
```
function Base.unsafe_convert(::Type{Ptr{Cfloat128}}, r::Base.RefValue{Float128})
    return Ptr{Cfloat128}(pointer_from_objref(r))
end
```
but I don't think that should be needed for pass-by-reference. (Pass-by-value is the can of worms for which we needed `Cfloat128`.)

amontoison:
Thanks @RalphAS for your quick answer!
I found the culprit in the Fortran code of `CUTEst`, a macro was needed to correctly compile the shared library in quadruple precision:
https://github.com/ralna/CUTEst/pull/62

All tests passed locally.
I will upgrade CUTEst_jll.jl with Yggdrasil.

PR #423, opened 2024-10-10
 Update CUTEst.jl for CUTEst v2.3.0

amontoison:


test/coverage.jl
```diff
@@ -2,6 +2,8 @@
 function coverage_increase(nlp::CUTEstModel{T}) where {T}
   status = Cint[0]
   n, m = nlp.meta.nvar, nlp.meta.ncon
+  # class = Vector{Cchar}(undef, 30)
+  # CUTEst.classification(T, nlp.libsif, nlp.status, nlp.funit, class)
```
tmigot, line 6: What was wrong here @amontoison ? As the function is no longer tested it seems

amontoison: Nick added this routine with the release 2.3.0 but I have some issue to make it work in Julia.
I think that Fortran changes the Julia pointer and it leads to a seg fault but we can't do that.
Julia own the memory.

With this new release of CUTEst, we also added a C version for all routines with string (*name*) and it means that strring are null-terminated (`Ptr{Cchar}` -> `Cstring`).

tmigot: Oki, thanks! Could you make an issue for that?

amontoison: Let's wait that all the C tests  in CUTEst passed first:
https://github.com/ralna/CUTEst/pull/67

Issue #275, opened 2022-01-14
Try syntax `CUTEstModel("...") do nlp` to allow automatic closing of nlp


abelsiqueira:

amontoison:
Not needed anymore, each CUTEst model is independant and have its own finalizer.

tmigot:
ok, so we no longer need to do finalize(nlp) at the end?
If so, I suggest we create a new issue to update the tutorials.

amontoison:
No, we don't need that since the complete refactor of `CUTEst.jl` and the release 1.0.

amontoison:
If you update the tutorials, please remove `CUTEst.select`.

PR #456, opened 2025-06-28
nlp doc change

### LINKED ISSUE #442: Update tutorials using CUTEst
to remove the finalize at the end, see #275 , and use the new selector.
Check:
- [ ] CUTEst.jl tutorials
- [ ] https://jso.dev/tutorials/
- [ ] Other packages docs

arnavk23:
Removed outdated call to `finalize(nlp)` and recommended using the `CUTEstModel("PROBLEM") do nlp ... end` syntax for automatic cleanup as mentioned #275, thus this Closes #442.

- Also want to add that we don't need a tutorial.md when the contents of the file is an introduction video and it is already mentioned on the readme.md file.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -67,9 +67,10 @@ You can pass parameters to `sifdecoder` by providing additional arguments to `CU
 using CUTEst
 
 for nh in 50:50:200
-  nlp = CUTEstModel{Float64}("CHAIN", "-param", "NH=$nh")
-  println("nh = $nh, nnzh = $(nlp.meta.nnzh)")
-  finalize(nlp)
+  CUTEstModel{Float64}("CHAIN", "-param", "NH=$nh") do nlp
+    println("nh = $nh, nnzh = $(nlp.meta.nnzh)")
+    finalize(nlp)
+  end
 end
 ```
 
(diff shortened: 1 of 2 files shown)


docs/src/index.md
```diff
@@ -67,9 +67,10 @@ You can pass parameters to `sifdecoder` by providing additional arguments to `CU
 using CUTEst
 
 for nh in 50:50:200
-  nlp = CUTEstModel{Float64}("CHAIN", "-param", "NH=$nh")
-  println("nh = $nh, nnzh = $(nlp.meta.nnzh)")
-  finalize(nlp)
+  CUTEstModel{Float64}("CHAIN", "-param", "NH=$nh") do nlp
+    println("nh = $nh, nnzh = $(nlp.meta.nnzh)")
+    finalize(nlp)
```
amontoison, line 72: We should maybe add a comment to specify why we call `finalize` here.
What fo you think?

We don't want to wait the next call of the garbage collector in this case and be sure that we can overwrite the same library for a different `nh`.

arnavk23: Yeah I agree, what do you think about my comment on tutorial.md and can you also look through the failing check

amontoison: Yes but not the main documentation page "Home". Having it in both places is file, but a standalone documentation page is maybe too much.

arnavk23:
@amontoison I think now this will pass all the checks.

amontoison (approved):

PR #459, opened 2025-06-30
Update docstring in src/model.jl

### LINKED ISSUE #440: Coexistence of multiple CUTEst models
Each `CUTEstModel` is, in theory, independent, and we can create as many models as we want; however, there is a corner case.  
If all `CUTEstModel` instances are based on different problems, this is not an issue.  
If all `CUTEstModel` instances are based on the same problem but with different precisions, this is not an issue either.  
However, if we have `CUTEstModel` instances for the same problem and precision, but with variable parameters (such as size), we encounter an issue.

This is because the shared library generated after decoding the problem is suffixed only with the name of the problem and its precision.  
The user must define the names of the decoded SIF files and shared libraries so that they do not conflict.

arnavk23:
This PR adds documentation to clarify an important corner case when using multiple CUTEstModel instances.

This note helps users understand:
- why conflicts can happen
- when conflicts do not happen

No code change; documentation only.
Closes #440

### DIFF
--- a/src/model.jl
+++ b/src/model.jl
@@ -65,7 +65,6 @@ Each CUTEstModel is, in theory, independent, and we can create as many models as
 !!! warning
     If we have CUTEstModel instances for the same problem and precision, but with variable parameters (such as size), we encounter an issue
     due to the shared library generated after decoding the problem, suffixed only with the name of the problem and its precision.
-    The user must define the names of the decoded SIF files and shared libraries so that they do not conflict.
 
 ## Examples
 


arnavk23:
@amontoison Please review the changes made here.

amontoison (approved):

src/model.jl
```diff
@@ -56,6 +56,19 @@ Finalize the current model by calling `finalize(nlp)` to avoid conflicts.
...
+## Coexistance of multiple CUTEst models
+
+Each CUTEstModel is, in theory, independent, and we can create as many models as we want : 
+- all CUTEstModel instances are based on different problems
+- all CUTEstModel instances are based on the same problem but with different precisions
+
+!!! warning
+    If we have CUTEstModel instances for the same problem and precision, but with variable parameters (such as size), we encounter an issue
+    due to the shared library generated after decoding the problem, suffixed only with the name of the problem and its precision.
+    The user must define the names of the decoded SIF files and shared libraries so that they do not conflict.
```
amontoison, line 68:
```suggestion
```

amontoison:
I removed the last sentence because we don't provide (yet) a way to specify a custom "tag" when we create the CUTEstModel.

PR #472, opened 2025-10-01
Update CUTEst.jl for SIFDecode 3.1.0

amontoison:
@jfowkes @nimgould
The new release of SIFDecode broke `CUTEst.jl` because `STOP 0` is written in `stderr` and not `stdout`.
I was checking that `stderr` was empty before.
This PR quickly fixes the issue.

It could be relevant for you to know that if you want to use it in the Python interface.

### DIFF
--- a/src/sifdecoder.jl
+++ b/src/sifdecoder.jl
@@ -139,7 +139,7 @@ function sifdecoder(
       ),
     )
     error_str = read(errlog, String)
-    if length(error_str) > 0
+    if length(error_str) > 0 && (error_str != "STOP 0\n")
       println(error_str)
       error("Unable to compile a shared library for the problem $name.")
     end


jfowkes:
Thanks @amontoison, this is why it's so annoying that gfortran added support for QUIET so late.

It doesn't seem to be an issue for the Python interface as it checks the return code.

nimgould:
The fortran standard is rather vague about this, but I now believe

STOP

gives status = 0, but does not output a message

STOP code

gives status = exit, written to standard error. Thus I propose that we have

   IF ( status == 0 ) THEN
    STOP
  ELSE
    WRITE( out,                                                         &
     "( /, ' Error exit from decoding stage. terminating execution.' )" )
    STOP status
  END IF

Any feelings or objections? 

https://www.cita.utoronto.ca/~merz/intel_f10b/main_for/mergedProjects/lref_for/source_files/rfstop.htm#:~:text=If%20stop%2Dcode%20is%20specified,to%20stop%2Dcode%20is%20returned.

nimgould:
Another thing, although we pass a code using STOP, it is equally valid to pass a string (which is then echoed). Indeed, I suspect that the code is actually passed as a string ... I'm not suggesting we do, but it would be more friendly than simply getting a final "STOP 3" message.

jfowkes:
So long as the return codes don't change I'm happy.

nimgould:
I presume that the MACOS actions machine is down, as all MAC jobs time out waiting for a server. Is this the US Gov shutdown in action?

amontoison:
> The fortran standard is rather vague about this, but I now believe
> 
> STOP
> 
> gives status = 0, but does not output a message
> 
> STOP code
> 
> gives status = exit, written to standard error. Thus I propose that we have
> 
> IF ( status == 0 ) THEN STOP ELSE WRITE( out, & "( /, ' Error exit from decoding stage. terminating execution.' )" ) STOP status END IF
> 
> Any feelings or objections?
> 
> https://www.cita.utoronto.ca/~merz/intel_f10b/main_for/mergedProjects/lref_for/source_files/rfstop.htm#:~:text=If%20stop%2Dcode%20is%20specified,to%20stop%2Dcode%20is%20returned.

I approve theses modifications 👍

